PR #2, opened 2022-10-27
Initial TS-7970 Support

ts-kris:
Initial support for all TS-7970 variants

* Includes in-FPGA MAX3100 support with multiple UARTs on a single chip-select which has had some of the locking reworked.
* Fully uses Etnaviv GPU driver. Can achieve 3D and 2D hardware acceleration in generic distributions.
* Uses open-source ath10k driver and publicly available firmware for Silex SX-SDMAC Wi-Fi/Bt
* Devicetree updated to follow modern syntax
* Single devicetree able to support both LSR and Silex Wi-Fi modules rather than having to select specific devicetree version.


ts-kris:
> **Squash [99f7df9](https://github.com/embeddedTS/linux-lts/commit/99f7df9689124e14c993d6a16edf294b67bfca44) in to [aaea58a](https://github.com/embeddedTS/linux-lts/commit/aaea58aeb52074ee34ab944bd5ef9f813188de41) ([98e4d4f](https://github.com/embeddedTS/linux-lts/commit/98e4d4f172cd7c65739c8c6201963541211476fb) may need to be touched too) before merging!**

Note that, those using spi-gpio may instead need num_chipselects according to devicetree bindings docs.

ts-kris:
@markfeathers added a bunch of upstream commits to the RTC and then added in additional functionality including:
* Control of temp sensing on battery (defaults to what upstream driver used to do)
* Setting of 75/85 battery % thresholds (default to what upstream driver used to do)
* Connecting the RTC temp sensor to HWMON
* Control of RTC offset for our emulated ISL12022 (real device unaffected as far as I can test)

I went back and tested/cleaned up all of the above (save for RTC offset, @markfeathers hammered on that exclusively), and integrated the TS-7970 DT and tsimx6_defconfig to support these.

Since I touched a whole bunch, @markfeathers can you please review?

@lionel, @luckshot, @randytsi369, Can at least one of your give this a review so we can get it all merged in? Thanks!

ts-kris:
Fixed some checkpatch issues I forgot were introduced along the way.

markfeathers (approved):

PR #4, opened 2022-11-29
Initial TS-7250-V3 support

markfeathers:


arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+
+		linux,cma {
+			compatible = "shared-dma-pool";
+			reusable;
+			size = <0x14000000>;
+			linux,cma-default;
+		};
+	};
+
+	regulators {
```
ts-kris: Regulator nodes should be one level higher, the `regulator` sub node is not needed and can be detrimental. i.e. remove `regulators { ... } and move the actual regulator nodes up one level.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+			regulator-max-microvolt = <1400000>;
+			regulator-name = "gpio_dvfs";
+			regulator-type = "voltage";
+			gpios = <&gpio5 3 GPIO_ACTIVE_HIGH>;
+			states = <1300000 0x1 1400000 0x0>;
+		};
+
+		reg_3v3: rev_3v3 {
+			compatible = "regulator-fixed";
+			regulator-name = "reg-3v3";
```
ts-kris: I've been trying to keep regulator names all caps like GPIO (which you have below) and organized by system and then function. e.g. this should be `3V3` the below ADC should be `ADC_VREF`, the above gpio should be `GPIO_DVFS`, etc.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+		compatible = "technologic,ts7100-wdt";
+		/* 5min timeout default, in case of slow userspace.
+		 * Set to 0 to disable WDT at startup. Userspace procs
+		 * can still start feeding later.
+		 */
+		timeout-sec = <300>;
+		reg = <0x54>;
+	};
+
+	rtc: m41t00s@68 {
```
ts-kris: Node and label are backwards, should be `m41t00s: rtc@68` Similar with gyro below, should be `ism1330: gyrometer@6a` Ensure this is the same through the whole device tree.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+&adc1 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_adc1>;
+	vref-supply = <&vref_adc_2v5>;
+};
+
+&can1 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_flexcan1>;
+	xceiver-supply = <&reg_3v3>;
```
ts-kris: This is powered from SW_5V not 3.3 net. Either remove this property (since SW_5V is basically system power) or fix it to be the correct voltage. Same with can2

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+};
+
+&can2 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_flexcan2>;
+	xceiver-supply = <&reg_3v3>;
+};
+
+&cpu0 {
```
ts-kris: `arm-supply` and `soc-supply` provided by imx6ul.dtsi and not needed here. I'm not sure if `dc-supply` is a valid binding or not. Is this a holdover from NXP kernel work?

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+	arm-supply = <&reg_arm>;
+	soc-supply = <&reg_soc>;
+	dc-supply = <&reg_gpio_dvfs>;
+};
+
+&ecspi3 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_ecspi3>;
+	fsl,spi-num-chipselects = <1>;
```
ts-kris: This should be `num-cs`, no longer uses the fsl specific property name.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+	cs-gpios = <&gpio4 12 0>;
```
ts-kris: Change the `0` to `GPIO_ACTIVE_LOW` as per our discussion on CS pins needing the active low polarity (or the active high property if they are active high).

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+
+	wilc_spi@0 {
```
ts-kris: Should be `wilc3000: wifi@0`

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+		compatible = "microchip,wilc3000", "microchip,wilc1000";
```
ts-kris: Note that the in-kernel driver is only for wilc1000 and last I tested won't work with this. Be sure that this is set up to work with the external wilc3000 module build (it looks like it but I havn't verified).

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+&fec2 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_enet2>;
+	phy-mode = "rmii";
+	phy-handle = <&ethphy1>;
+	phy-supply = <&reg_3v3>;
+	status = "okay";
+};
+
+&gpc {
```
ts-kris: Double check these, I'm 90% sure these were from NXP specific kernel setup and I do not believe these are used or matter in vanilla kernel i.MX6UL support.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+		};
+	};
+};
+
+&i2c1 {
+	clock-frequency = <100000>;
+	pinctrl-names = "default", "gpio";
+	pinctrl-0 = <&pinctrl_i2c1>;
+	pinctrl-1 = <&pinctrl_i2c1_gpio>;
+	scl-gpios = <&gpio1 2 GPIO_ACTIVE_HIGH>;
```
ts-kris: These need open drain property, e.g. `(GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)`

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+	sda-gpios = <&gpio1 3 GPIO_ACTIVE_HIGH>;
+	status = "okay";
+
+	magnet: magnetometer@1e {
```
ts-kris, line 123: Do we have two magnetometers on this?

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+		"",
+		"";
+};
+
+&i2c3 {
+	clock-frequency = <100000>;
+	pinctrl-names = "default", "gpio";
+	pinctrl-0 = <&pinctrl_i2c3>;
+	pinctrl-1 = <&pinctrl_i2c3_gpio>;
+	scl-gpios = <&gpio3 06 GPIO_ACTIVE_HIGH>;
```
ts-kris: These should have open drain marked, see above.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+&uart1 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_uart1>;
+	status = "okay";
+};
+
+&uart3 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_uart3>;
+	fsl,uart-has-rtscts;
```
ts-kris: This is just `uart-has-rtscts` the fsl specific property is no longer used. Same issue below in uart4

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+				MX6UL_PAD_CSI_DATA00__EIM_AD00		0x1b069
+			>;
+		};
+	};
+};
+
+&usbotg1 {
+	dr_mode = "host";
+	disable-over-current;
+	vbus-supply = <&reg_3v3>;
```
ts-kris: Is 3.3 V correct for USB VBUS supply?

ts-kris:
@markfeathers See comments on dts. I don't know enough about the other subsystems to be able to comment as deeply but I'll still look through them when I can and see if anything jumps out.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1200 @@
...
+		timeout-sec = <300>;
+		reg = <0x54>;
+	};
+
+	rtc: m41t00s@68 {
+		compatible = "m41t00";
+		reg = <0x68>;
+	};
+
+	gyro: ism1330@6a {
```
lionel: Typo: 'ism330', not `ism1330`.

arch/arm/boot/dts/imx6ul-ts7250v3.dts
```diff
@@ -0,0 +1,1155 @@
...
+				reg = <1>;
+				spi-max-frequency = <19800000>;
+				size = <0x800>;
+				address-width = <16>;
+				pagesize = <64>;
+			};
+
+			/* Onboard FPGA Flash */
+			is25lq016b: flash@2 {
+				compatible = "issi,is25lq016b", "jedec,spi-nor";
```
ts-kris: Check the BOM. If we have multiple parts that we could sub in here, you can use just `compatible = "jedec,spi-nor"` since everything we use would have JEDEC IDs. Otherwise, the kernel is too smart for its own good and yells at you if it detects a part different than what you specified if we end up doing a build with a compatible part from a different manufacturer.

ts-kris (dismissed):
Reviewed, not tested.

ts-kris (dismissed):
Thumbed quickly through the defconfig, looks good to me, I did not attempt a boot. Re-approved.

drivers/irqchip/irq-ts71xxweim.c
```diff
@@ -0,0 +1,167 @@
+// SPDX-License-Identifier: GPL-2.0
+
```
luckshot, line 2: WEIM = Wireless External Interface Module

luckshot (dismissed):

drivers/rtc/Kconfig
```diff
@@ -404,22 +404,13 @@ config RTC_DRV_ISL1208
 
 config RTC_DRV_ISL12022
```
ts-kris: Why were these lines removed? Removes REGMAP_I2C from the ISL driver, and removes the ISL HWMON driver.

drivers/power/reset/tssupervisor-reset.c
```diff
@@ -0,0 +1,161 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
ts-kris, line 1: Is there any reason to not implement this as a poweroff controller too? Since the uC presumably has power and reset control of the whole board, it would be nice to have a poweroff mode too.

arch/arm/boot/dts/imx6ul-ts7250v3.dtsi
```diff
@@ -0,0 +1,1030 @@
...
+	status = "okay";
+
+	magnet: magnetometer@1e {
+		compatible = "st,lis2mdl";
+		reg = <0x1e>;
+		interrupt-parent = <&gpio1>;
+		interrupts = <14 IRQ_TYPE_LEVEL_HIGH>;
+	};
+
+	legacy_supervisor: supervisor@54 {
```
ts-kris: If the only support of this legacy supervisor is WDT, and newer supervisor _doesn't_ have WDT support, this should be labeled as a watchdog.

arch/arm/boot/dts/imx6ul-ts7250v3.dtsi
```diff
@@ -0,0 +1,1030 @@
...
+			reg = <0x100 32>;
+			interrupt-parent = <&fpga_intc>;
+			interrupts = <9>;
+			clocks = <&fpga_clk_weim_bclk>;
+			clock-names = "spi-oc-clk";
+			opencores-spi,idx = <0>;
+			opencores-spi,num-chipselects = <3>;
+
+			dioheader: spi@0 {
+				compatible = "technologic,spidev";
```
ts-kris: I don't think we added an exception to spidev.c, in 5.10 this can safely be left as "spidev" it will give a warning but the bus will still function.

ts-kris (commented):
A few things out of place in the dts, which, pending discussion may be able to be left alone. See other comments. I'm getting ready to do a boot test of this here.

drivers/power/reset/tssupervisor-reset.c
ts-kris: Ignore me, I missed this was also a poweroff driver.

ts-kris (changes requested):
Both minimal and regular `tsimx6ul_defconfig` should have:
CONFIG_EXTRA_FIRMWARE="imx/sdma/sdma-imx6q.bin"
CONFIG_EXTRA_FIRMWARE_DIR="firmware"

SDMA firmware should be in-kernel since many things use it and I'm not sure if its safe to load after some peripherals have started.

ts-kris:
> Both minimal and regular `tsimx6ul_defconfig` should have...

Note that, this should also be in `tsimx6_*` once we get those separated back out.

ts-kris (changes requested):
Magnetometer seems to not show up correctly, I've verified it is present on my unit:
```
[   45.158306] st-magn-i2c 0-001e: supply vdd not found, using dummy regulator
[   45.160717] st-magn-i2c 0-001e: supply vddio not found, using dummy regulator
[   45.330594] iio iio:device7: DRDY on pdata not valid.
[   45.562887] st-magn-i2c: probe of 0-001e failed with error -22
```

arch/arm/boot/dts/imx6ul-ts7250v3.dtsi
```diff
@@ -0,0 +1,1030 @@
...
+&gpio2 {
+	gpio-line-names =
+		"ENET1_RX_DATA0", "ENET1_RX_DATA1", "ENET1_RX_EN",
+		"ENET1_TX_DATA0", "ENET1_TX_DATA1", "ENET1_TX_EN",
+		"ENET1_TX_CLK", "ENET1_RX_ER", "ENET2_RX_DATA0",
+		"ENET2_RX_DATA1", "ENET2_RX_EN", "ENET2_TX_DATA0",
+		"ENET2_TX_DATA1", "ENET2_TX_EN", "ENET2_TX_CLK",
+		"ENET2_RX_ER", "EMMC_CMD", "EMMC_CLK", "EMMC_D0", "EMMC_D1",
+		"EMMC_D2", "EMMC_D3", "", "", "", "", "", "", "", "", "", "";
+};
```
ts-kris: For future reference, I don't think its necessary, and it actually may be detrimental, to name pins as GPIO which generally cannot be GPIO. e.g. all of the ENET pins only go to the PHY, and while they could be put in to GPIO mode, are not usable as GPIO normally on this system. Saves us time in labeling and reduces noise in `gpioinfo` to end users.

ts-kris (changes requested):
Both `tsimx6_*` defconfigs should include:
CONFIG_RTC_DRV_ISL12022=y (or m)
RTC_DRV_ISL12022_HWMON=y (or m)
(but both must be y or m I believe)

This should also be the case when we split out `tsimx6_*`

ts-kris:
I've otherwise poked at a few other peripherals, Wi-Fi, IIO that were present, SD, eMMC, eth0 (no eth1 on my board), random DIO pins, RTC, USB. Everything else I poked seemed functional.

ts-kris (dismissed):
Boot tested, everything looks good overall. Not every peripheral was tested, but, as indicated previously I did random testing of DIO pins, networking, UART, USB, IIO, etc.

ts-kris:
Just adding a note here that, with this much of a change, we should probably tag the final merge commit of this as 5.10.170.1-ts

drivers/mfd/tssupervisor-core.c
```diff
@@ -0,0 +1,246 @@
...
+		dev_err(dev, "error reading reg %u", SUPER_MODEL);
+	err = regmap_read(super->regmap, SUPER_REV_INFO, &revision);
+	if (err < 0)
+		dev_err(dev, "error reading reg %u", SUPER_REV_INFO);
+	dev_info(&client->dev, "Model %04X rev %d%s\n",
+		 model,
+		 revision & 0x7fff,
+		 revision & 0x8000 ? " (DIRTY)" : "");
+
+	if (model == MODEL_TS_7250_V3) {
```
lionel, line 206: What is the reason for locking this driver to the 7250v3? (or at least its creation of sysfs entries ... otherwise succeeding but not creating them?). We might want a more maintainable (and generic) way to make these determinations while at the same time supporting a lot of new models that won't ID as 7250.

drivers/mfd/Kconfig
```diff
@@ -1794,6 +1794,16 @@ config MFD_TC6393XB
...
 
+config MFD_TS_SUPERVISOR
+	tristate "embeddedTS Supervisor"
+	select MFD_CORE
+	select REGMAP_I2C
+	depends on I2C && OF
+	help
+	  Support for embeddedTS supervisory controller.  This includes an adc
+	  driver for system rails, supercap backup management, and some misc
+	  controls.
```
lionel, line 1805: "Support for **the** embeddedTS" ...
Suggested clean-up idea:
```
Support for the embeddedTS supervisory microcontroller, which provides
some/all of: power/reset control, an RTC, supercaps, and (via IIO)
system rail voltage ADCs and a temperature sensor.
```

drivers/irqchip/irq-ts71xxweim.c
```diff
@@ -0,0 +1,167 @@
...
+	u32 mask;
+} priv;
+
+static const struct of_device_id tsweim_intc_of_match_table[] = {
+	{.compatible = "technologic,ts71xxweim-intc", },
+	{},
+};
+MODULE_DEVICE_TABLE(of, tsweim_intc_of_match_table);
+
+static void ts7120_intc_mask(struct irq_data *d)
```
lionel: What would happen if we just doubled-down on this being our WEIM-acccessible, irqchip for TS boards and get rid of the 7120 references?

lionel: An important thing to remember about this irqchip is that it is FPGA-based, and so WEIM and its IRQs need to be off-limits when updating the FPGA code.

drivers/power/reset/tssupervisor-reset.c
```diff
@@ -0,0 +1,161 @@
...
+#include <linux/regmap.h>
+#include <linux/mfd/core.h>
+#include <linux/reboot.h>
+#include <linux/mfd/ts_supervisor.h>
+
+/* We need a static device to support this for shutdown/reboot hooks */
+static struct device *ts_rstc_device;
+static atomic_t ts_restart_nb_refcnt = ATOMIC_INIT(0);
+
+static ssize_t reboot_reason_show(struct device *dev,
```
lionel, line 15: Ugh, I'm missing how this is getting called. Is there a macro that generates a reference to the symbol `reboot_reason_show`?

drivers/power/reset/Kconfig
```diff
@@ -190,6 +190,12 @@ config POWER_RESET_ST
 	help
 	  Reset support for STMicroelectronics boards.
 
+config POWER_RESET_TS_SUPERVISOR
+	bool "embeddedTS Supervisory reset driver"
+	depends on MFD_TS_SUPERVISOR
+	help
+	  Reset support for embeddedTS boards
```
lionel, line 197: vs. a CPU-generated reset? At one point we were experimenting with having the CPU handle the reset. Is leaving this out all there is to it ?-)

drivers/power/reset/tssupervisor-reset.c
lionel: After `reboot_reason_show()` is `ATTR_RO(reboot_reason)`; `ATTR_RO(name)` -> Generates a `struct device_attr` populated by `__ATTR_RO(name)` -> that has an entry referencing _name_`_show`. This is the preferred way to generate these tables/entries for `sysfs`.

lionel (dismissed):
Not tested, but PR is fine as-is.

drivers/power/reset/Kconfig
markfeathers: The CPU can handle the reset via the cpu wdog interface, but when the microcontroller is told to reset directly it tracks the reboot reason.

ts-kris (dismissed):
I discussed the latest changes with Mark and agree with the general idea behind them to clean up some of the numbering and making things more generic by name. Changes not tested.

drivers/irqchip/irq-ts71xxweim.c
```diff
@@ -0,0 +1,167 @@
...
+		pr_err("%s: unable to add irq domain\n", np->name);
+		return -ENOMEM;
+	}
+
+	irq_set_handler_data(priv.irq, &priv);
+	irq_set_chained_handler(priv.irq, tsweim_irq_handler);
+
+	platform_set_drvdata(pdev, &priv);
+
+	pr_info("ts7120-intc IRQ domain created\n");
```
lionel: Oops, not `ts7120-intc` any more. Compatible string is `ts71xxweim-intc`. Maybe match that?

markfeathers:
Fixed one remaining 7120 reference in the IRQ driver, and fixed some minor whitespace issues. Did a build check.

lionel (approved):

ts-kris (approved):

PR #7, opened 2023-01-06
Initial TS-7990 support

ts-kris:
Add support for TS-7990 and variants.

Also includes some driver patches.
- The pixcir driver (LCD cap touch) needed modification to put it in active high IRQ mode. Patch was based on similar modification in linux-tsimx repo.
- The ads7846 driver (Okaya/Microtip resistive touch) did not have an SPI ID struct and therefore fails to auto-load as a module. This will get pushed upstream shortly hopefully but likely won't get backported here so that commit will need to hang around.

As the TS-7990 uses the WILC3000 Wi-Fi module, the defconfig was modified to not enable 802.11 power savings by default as this has been observed to cause issues with the driver.

**NOTE**
This should be merged after #5, in either case there will likely be some conflicts with dts/Makefile that need to be sorted out.

The only variants tested were Quad LXD Rev B, and Quad Okaya Rev B. At this time I do not have any Solo CPUs nor Rev A PCBs.

### DIFF
--- a/arch/arm/boot/dts/Makefile
+++ b/arch/arm/boot/dts/Makefile
@@ -472,6 +472,12 @@ dtb-$(CONFIG_SOC_IMX6Q) += \
 	imx6dl-ts4900-2.dtb \
 	imx6dl-ts4900-14.dtb \
 	imx6dl-ts7970.dtb \
+	imx6dl-ts7990-lxd-revb.dtb \
+	imx6dl-ts7990-lxd.dtb \
+	imx6dl-ts7990-microtips-revb.dtb \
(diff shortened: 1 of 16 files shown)


luckshot (dismissed):

ts-kris:
As of #9, this PR needs some rework to remove conflicts

ts-kris:
Dropped commits to tsimx6_defconfig and verified that ts_defconfig had those options.

ts-kris:
Latest force push was to touch compatibility strings to change from ts4900 to ts7990 across all files.

markfeathers (approved):

luckshot (approved):

PR #8, opened 2023-01-18
Initial TS-7553-V2 support

ts-kris:
Support for TS-7553-V2 in 5.10

This adds its own defconfig with the intent that it will likely be merged in to a more singular defconfig in the future. However, to ensure a point in time where everything is tested and support, that is included in this PR as well.

This also adds support for the LCD as a framebuffer as it was in older kernels. Long term, the goal is to be able to support the monochrome LCD as a DRM supported panel with backlight, etc. The goal for this PR however is basic support of all features.

Additionally, it has been observed that microSD access is a bit slow here. Some optimizations and improvements can come later to speed that up.


randytsi369 (dismissed):
No issues found during review. Not tested.

ts-kris:
Touches ts_defconfig to do the following:
- Move many things from builtins to modules, e.g. touchscreen drivers (note that this needs commit 25c6e79802a6d5e929d0691ee138f393513f8721 to work on resistive TPCs), audio codecs, IIO drivers, etc.
- Remove useless drivers, e.g. any non PCIe or USB network device driver that is not used on any of our platforms, filesystem drivers for ancient/unrelated devices, NAND specific drivers, regulator drivers, some media drivers, etc.
- Add in many new features from tsimx6ul_defconfig in 4.9 back here as modules, e.g. ipv6 netfilter
- Remove framebuffer console, necessary for TS-7553-V2 use of mono LCD but this may break uses of TPC where console should be present on the FB. Also add the SNVS powerkey driver as a built in to maintain the TS-7553-V2 demo keypad input. This can be removed/modularized later once the mono LCD gets a DRM panel driver.
- Add in driver support for TS-7553-V2 IIO

I did testing of the defconfig changes along the way under TS-4900, TS-7970, and TS-7990. Everything works as expected but I have not re-run through all of the testing again after doing this final squash. The last testing I did was with framebuffer console enabled and SNVS power key as a module. I don't suspect any issues (aside from resistive TPCs as I noted above) but will re-test things tomorrow.

markfeathers (approved):

luckshot (approved):

PR #9, opened 2023-01-18
Linux 5.10.y common config

markfeathers:
Create initial ts_defconfig with base 6ul support


ts-kris (approved):
I did force push moving a driver back to a module that this inadvertently changed to a builtin.

Otherwise, tested on TS-7970, quad/solo, rev E, F, and H (quad only), TS-4900 w/ TS-8950 and TS-8390

lionel (commented):
So we want `CONFIG_NFS_V4=m` but `CONFIG_NFS_V4_[12]*` to be `y`?

Nevermind. NFS_V4 is tri-state, but 4_1 and 4_2 are boolean settings. I had been thinking that they'd be including modules in the kernel while the rest of NFS_V4 is in a module.

ts-kris:
> So we want `CONFIG_NFS_V4=m` but `CONFIG_NFS_V4_[12]*` to be `y`?

CONFIG_NFS_V4_1/_2 are feature options to NFSV4:
```
  │ │                  <M>     NFS client support for NFS version 4                                            │ │  
  │ │                  [ ]     Provide swap over NFS support                                                   │ │  
  │ │                  [*]   NFS client support for NFSv4.1                                                    │ │  
  │ │                  [*]     NFS client support for NFSv4.2
```

EDIT: the tabbing is weird, even in menuconfig, but, you can't have V4_1 without V4 being enabled:
```
  │   Prompt: NFS client support for NFSv4.1
  │   Depends on: NETWORK_FILESYSTEMS [=y] && NFS_V4 [=m]

lionel (approved):
Looks like a good starting point for a common defconfig as additional products are merged in.

PR #18, opened 2023-03-17
Linux 5.10.y add ts7800v2

luckshot:


markfeathers (dismissed):

drivers/gpio/gpio-ts7800-v2.c
```diff
@@ -0,0 +1,505 @@
...
+		reg |= bit;
+	else
+		reg &= ~bit;
+
+	writel(reg, priv->syscon + reg_num);
+	priv->direction[offset / 32] &= ~(1 << offset % 32);
+	spin_unlock_irqrestore(&priv->lock, flags);
+
+	return ret;
+}
```
ts-kris: This could potentially be an issue as it sets direction before setting value. The correct thing in hardware is to set the value and then direction. That way, when the direction is changed, the output pin is reflecting the state that is expected of it. If direction is set first, the output value might be an invalid/unexpected state for a brief moment before the output value is changed.

drivers/gpio/gpio-ts7800-v2.c
```diff
@@ -0,0 +1,505 @@
...
+		return -ENOMEM;
+
+	priv->syscon = membase;
+
+	memset(priv->direction, 0xFF, sizeof(priv->direction));
+	memset(priv->ovalue, 0, sizeof(priv->ovalue));
+	/* Set all the DIO/LCD outputs high (they are open-drain) */
+	/* ... Except lcd_04.  That one drives 3.3V. Set it low.  */
+	reg = readl(priv->syscon + 8) | 0x3ff7ffff;
+	writel(reg, priv->syscon + 8);
```
ts-kris: I believe it is bad form to have a GPIO driver set GPIO pin states. If these need to be set then they should be set in bootloader.

drivers/gpio/gpio-ts7800-v2.c
```diff
@@ -0,0 +1,505 @@
...
+	unsigned int offset)
+{
+	struct ts7800v2_gpio_priv *priv = to_gpio_ts7800v2(chip);
+
+	if (offset >= TS7800V2_NR_DIO)
+		return -EINVAL;
+
+	return !!(priv->direction[offset / 32] & (1 << offset % 32));
+
+}
```
ts-kris: This is a problem as-is. See detailed comments on line 456-457.

drivers/gpio/gpio-ts7800-v2.c
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	priv = devm_kzalloc(dev, sizeof(*priv), GFP_KERNEL);
+	if (!priv)
+		return -ENOMEM;
+
+	priv->syscon = membase;
+
+	memset(priv->direction, 0xFF, sizeof(priv->direction));
+	memset(priv->ovalue, 0, sizeof(priv->ovalue));
```
ts-kris: These lines make a naive assumption of the state of GPIO. In combination with `gpio_get_direction` the driver could potentially return invalid state data if the hardware is probed directly (e.g. by bootloader).

I think the correct fix of this is to completely remove `priv->direction` and `priv->ovalue`, the two are only ever written to with the exception of `priv->direction` being returned in `gpio_get_direction`. I think if `gpio_get_direction` is refactored to poll the hardware, then `priv->direction` and `priv->ovalue` could be completely removed.

Unless I'm missing a transaction, every read/write I see uses only a single 32-bit value as well as the address offset set up in `priv->syscon`.

drivers/gpio/gpio-ts7800-v2.c
ts-kris: It might also make sense to have this function call `gpio_set`. This would end up making `direction_output` calls have a longer path, but, would keep the two tasks segmented and easier to debug.

drivers/gpio/gpio-ts7820.c
```diff
@@ -0,0 +1,237 @@
...
+
+static int gpio_ts7820_irq_set_type(struct irq_data *d, unsigned int type)
+{
+	struct gpio_chip *gc = irq_data_get_irq_chip_data(d);
+	struct ts7820_gpio *ts7820_gpio = gpiochip_get_data(gc);
+	unsigned int hwirq = irqd_to_hwirq(d);
+	int ret = 0;
+
+	irq_set_status_flags(d->irq, IRQ_IS_POLLED);
+	return 0;
```
ts-kris: Is this breaking IRQs? Why is it just returning 0? Why is there code after it that can never be reached?

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_pins &spi0_cs_gpio_pins>;
+	#address-cells = <1>;
+	#size-cells = <0>;
+	status = "okay";
+	num-cs = <5>;
+	cs-gpios = <0>, <0>, <0>, <0>, <&gpio0 28 GPIO_ACTIVE_HIGH>;
```
ts-kris: Should be `GPIO_ACTIVE_LOW`

ts-kris: If it really is active high, I think another flag needs to be added in the node parameters to indicate it as such.

drivers/watchdog/ts7100_wdt.c
```diff
@@ -0,0 +1,241 @@
...
+	if (!of_property_read_u32(client->dev.of_node, "timeout-sec", &t))
+		if (!watchdog_timeout_invalid(&ts_wdt_wdd, t))
+			ts_wdt_wdd.timeout = t;
+
+	/*
+	 * Set the timeout (most likely from DTS) and do a single feed until
+	 * userspace takes over finally.
+	 */
+	ts7100_wdt_set_timeout(&ts_wdt_wdd, ts_wdt_wdd.timeout);
+	ts7100_wdt_start(&ts_wdt_wdd);
```
ts-kris: Do we want to refactor this to be more like that changes made to ts_wdt? https://github.com/embeddedTS/linux-lts/blob/linux-5.10.y/drivers/watchdog/ts_wdt.c#L196-L216

Essentially, we don't enable self feeding, or even do a single feed at startup unless `enable-early` flag is set in FDT. The `enable-early` flag is not a standard kernel mechanism, but, a couple drivers use a similar notation.

General WDT paradigm is that the WDT is either running or not before the kernel starts. There is a kernel config option, CONFIG_WATCHDOG_HANDLE_BOOT_ENABLED, that if set _and_ the WDT hardware is running (it is up to the driver to check if the WDT is running and set `WDOG_HW_RUNNING` if it is), the kernel will feed until userspace takes over. Otherwise, it is assumed that the kernel needs to do nothing and let userspace take control.

drivers/irqchip/irq-ts7840.c
```diff
@@ -0,0 +1,155 @@
...
+	writel(reg & ~(1 << d->hwirq), data->base + IRQ_MASK);
+}
+
+static void ts7840_irq_unmask(struct irq_data *d)
+{
+	struct ts7840_irq_data *data = irq_data_get_irq_chip_data(d);
+	u32 reg = readl(data->base + IRQ_MASK);
+
+	writel(reg | (1 << d->hwirq), data->base + IRQ_MASK);
+}
```
ts-kris: Is there a need in RMW functions like this to have a spinlock?

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
+// SPDX-License-Identifier: (GPL-2.0 OR MIT)
+/*
+ * Device Tree file for embeddedTS TS-7800-V2
+ * Copyright (C) 2017-2022 embeddedts Systems, Inc. dba embeddedTS
```
ts-kris: "embeddedts Systems"

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
+// SPDX-License-Identifier: (GPL-2.0+ OR X11)
+/*
+ * Device Tree file for embeddedTS TS-7840
+ * Copyright (C) 2017-2022 embeddedts Systems, Inc. dba embeddedTS
```
ts-kris: "embeddedts Systems"

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+
+		isa-io8@2000000 {
+			compatible = "simple-bus";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			ranges = <0 3 0x2000000 0x1000000>;
+		};
+
+		isa-io16@2000000 {
```
ts-kris: I believe the correct naming for these should be something like:
`isa-mem8: bus@0`
`isa-mem16: bus@0`
`isa-io8: bus@200000`
`isa-io16: bus@200000`

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+			compatible = "technologic,ts7800v2-gpio";
+			gpio-controller;
+			#gpio-cells = <2>;
+			base = <64>;
+			ngpios = <118>;
+			#address-cells = <0x1>;
+			#size-cells = <0>;
+			reg = <2 0x0 0x60>;
+
+			gpio-line-names=
```
ts-kris: Prefer a more compact format for `gpio-line-names` (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+	status = "okay";
+	phy = <&phy0>;
+	phy-mode = "rgmii-id";
+	buffer-manager = <&bm>;
+	bm,pool-long = <0>;
+	bm,pool-short = <1>;
+};
+
+&gpio0 {
+	gpio-line-names =
```
ts-kris: Prefer a more compact format for `gpio-line-names` (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+		"",
+		"",
+		"",
+		"SPI_0_CS3#",
+		"GE_PHY_INT#",
+		"CPU_SPEED_1",
+		"CPU_SPEED_2";
+};
+
+&gpio1 {
```
ts-kris, line 347: Prefer a more compact format for `gpio-line-names` (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+		"";
+};
+
+&i2c0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&i2c0_pins>;
+	clock-frequency = <100000>;
+	status = "okay";
+
+	silabs_wdt: silabs-wdt@54 {
```
ts-kris: Should be `silabs_wdt: watchdog@54`

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+		compatible = "technologic,ts7100-wdt";
+		reg = <0x54>;
+	};
+
+	rtc1: m41t00s@68 {
```
ts-kris: Should be `m41t00s: rtc@68`

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+		compatible = "m41t00";
+		reg = <0x68>;
+	};
+
+	accelerometer_mm8451: accelerometer@1c {
```
ts-kris: Should be `mma8451: accelerometer@1c`

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+		"",
+		"",
+		"",
+		"";
+};
+
+&i2c0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&i2c0_pins>;
+	clock-frequency = <100000>;
```
ts-kris, line 358: Is there any specific reason to use 100 kHz bus?

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+	#size-cells = <0>;
+	status = "okay";
+	num-cs = <5>;
+	cs-gpios = <0>, <0>, <0>, <0>, <&gpio0 28 GPIO_ACTIVE_HIGH>;
+
+	offboard_flash: flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <1000000>;
+	};
```
ts-kris, line 451: The name is offboard flash, that also coincides with the error I see on bootup:
```
[    1.885140] spi-nor spi0.0: unrecognized JEDEC id bytes: ff ff ff ff ff ff
[    1.892115] spi-nor: probe of spi0.0 failed with error -2
```

Is this something we use in production or is it completely unused? If the latter, this should probably be removed from FDT as it is a dead device.

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+
+	wilc_spi@3 {
```
ts-kris: Should be `wilc: wifi@3`

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+		compatible = "microchip,wilc3000", "microchip,wilc1000";
```
ts-kris: Drop the `microchip,wilc1000` compatible. Not actually compatible with the upstream, in-kernel, WILC driver.

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	model = "embeddedTS TS-7800-V2";
+	compatible = "embeddedts,a385-ts7800-v2", "marvell,armada385", "marvell,armada380";
+
+	aliases {
+		ethernet1 = &eth0;
```
ts-kris: Should this be `ethernet0`? Does this device offer two ports?

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+#include <dt-bindings/interrupt-controller/irq.h>
+
+/ {
+	model = "embeddedTS TS-7840";
+	compatible = "technologic,a385-ts7840", "marvell,armada385", "marvell,armada380";
+
+	aliases {
+		ethernet1 = &eth0;
+		ethernet2 = &eth1;
+		ethernet3 = &eth2;
```
ts-kris: Shouldn't this start at `ethernet0`?

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		#size-cells = <0>;
+	};
+
+	aux_i2c: aux-i2c {
+		compatible = "i2c-gpio";
+		gpios = <&fpga_bank1 25 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>, /* sda */
+			<&fpga_bank1 26 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>; /* scl */
+		#address-cells = <1>;
+		#size-cells = <0>;
+	};
```
ts-kris: Couple of things on all of these:
1) Prefer label and node format similar to `<blah>_i2c: i2c@<number>` where number can just be 0 indexed most likely unless there are other nodes of i2c starting at 0
2) Are there by default any devices connected to these that we would want to attach a driver to?

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+
+	leds {
+		compatible = "gpio-leds";
+
+		led-0 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_POWER;
+			gpios = <&fpga_bank0 19 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "heartbeat";
```
ts-kris: I think this would be the only board we have green LED set up as a heartbeat. Not an issue either way, but, it begs the question, do we want to have this match other existing platforms or do we want other platforms matching this?

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+
+	soc {
+		ranges = <MBUS_ID(0xf0, 0x01) 0 0xf1000000 0x100000
+			  MBUS_ID(0x01, 0x1d) 0 0xfff00000 0x100000
+			  MBUS_ID(0x09, 0x19) 0 0xf1100000 0x10000
+			  MBUS_ID(0x09, 0x15) 0 0xf1110000 0x10000
+			  MBUS_ID(0x0c, 0x04) 0 0xf1200000 0x100000>;
+	};
+
+	tsfpga_pcie: tsfpga-pcie {
```
ts-kris: Should be `tsfpga_pcie: pcie`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		compatible = "technologic,ts78xx-mfd";
+		#address-cells = <2>;
+		#size-cells = <1>;
+
+		vendor = <0x1e6d>;
+		device = <0x7840>;
+
+		/* <ranges> is set in the driver at runtime */
+
+		fpga_pll0c1: fpga_pll0c1 {
```
ts-kris: Devicetree spec doesn't have anything specifically for PLLs, but, I think the correct generic node name would likely either be `clock` or we can just called it `pll`? I'm not completely sure. e.g. `fpga_pll0c1: clock`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			compatible = "fixed-clock";
+			#clock-cells = <0>;
+			clock-frequency = <50000000>;
+		};
+
+		fpgarng@a4 {
```
ts-kris: Should be `fpga_rng: rng@a4` or similar.

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		fpgairqc: irq-controller@b0 {
+			compatible = "technologic,ts7840-irqc";
+			reg = <0 0xb0 4>;
+			interrupt-controller;
+			#interrupt-cells = <2>;
+			interrupt-parent = <&gpio1>;
+			interrupts = <3 IRQ_TYPE_LEVEL_HIGH>;
+		};
+
+		fpga_bank0: ts7820-gpio@24 {
```
ts-kris: Should be `fpga_bank0: gpio@24`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			interrupts = <16 IRQ_TYPE_LEVEL_HIGH>;
+
+			interrupt-controller;
+			#interrupt-cells = <2>;
+
+			gpio-controller;
+			#gpio-cells = <2>;
+			reg = <0 0x24 0x1c>;
+
+			gpio-line-names =
```
ts-kris: Prefer a more compact format for gpio-line-names (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+				"left_grn_led_padn",
+				"left_yel_led_padn",
+				"bt_en_pad",
+				"wifi_en_pad",
+				"prog_silab_clk_padn",
+				"prog_silab_data_pad",
+				"cpu_push_sw_padn";
+		};
+
+		fpga_bank1: ts7820-gpio@40 {
```
ts-kris: Should be `fpga_bank1: gpio@40`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			interrupts = <17 IRQ_TYPE_LEVEL_HIGH>;
+
+			interrupt-controller;
+			#interrupt-cells = <2>;
+
+			gpio-controller;
+			#gpio-cells = <2>;
+			reg = <0 0x40 0x1c>;
+
+			gpio-line-names =
```
ts-kris: Prefer a more compact format for gpio-line-names (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+				"aux_i2c_dat_pad",
+				"aux_i2c_clk_pad",
+				"",
+				"",
+				"",
+				"",
+				"";
+		};
+
+		fpga_bank2: ts7820-gpio@5c {
```
ts-kris, line 238: Same as above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			interrupts = <18 IRQ_TYPE_LEVEL_HIGH>;
+
+			interrupt-controller;
+			#interrupt-cells = <2>;
+
+			gpio-controller;
+			#gpio-cells = <2>;
+			reg = <0 0x5c 0x1c>;
+
+			gpio-line-names =
```
ts-kris: Same as above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+				"jp1_padn",
+				"jp2_padn",
+				"",
+				"",
+				"",
+				"";
+		};
+
+		/* COM1 RS-232 */
+		fpga_uart0: fpga-uart0 {
```
ts-kris: Should be `fpga_uart0: uart@0`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x100 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <0 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* COM2 RS-232 */
+		fpga_uart1: fpga-uart1 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x108 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <1 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* GPS */
+		fpga_uart2: fpga-uart2 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x110 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <2 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* Nimbelink */
+		fpga_uart3: fpga-uart3 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x118 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <3 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* Iridium Modem */
+		fpga_uart4: fpga-uart4 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x120 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <4 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* DSL Modem */
+		fpga_uart5: fpga-uart5 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x128 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <5 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* Mikrobus UART */
+		fpga_uart6: fpga-uart6 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x130 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <6 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* RS-485 port */
+		fpga_uart7: fpga-uart7 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x138 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <7 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* XBEE */
+		fpga_uart8: fpga-uart8 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x140 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <8 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* Spare */
+		fpga_uart9: fpga-uart9 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			reg = <0 0x148 8>;
+			interrupt-parent = <&fpgairqc>;
+			interrupts = <19 IRQ_TYPE_LEVEL_HIGH>;
+			clock-frequency = <1843200>;
+			reg-io-width = <1>;
+			fifo-size = <64>;
+		};
+
+		/* Spare */
+		fpga_uart10: fpga-uart10 {
```
ts-kris: See above

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		};
+
+		mikro_pwm: pwm@600 {
+			compatible = "technologic,pwm";
+			reg = <0 0x600 8>;
+			clocks = <&fpga_pll0c1>;
+			clock-names = "pwm-input-clk";
+		};
+
+		ts_mmc: ts_mmc@800 {
```
ts-kris: Should be `ts_mmc: mmc@800`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			compatible = "ts,tssdcore2";
+			reg = <0 0x800 32>;
+			//interrupt-parent = <&fpgairqc>;
+			//interrupts = <14 IRQ_TYPE_LEVEL_HIGH>;
+			busy-gpios = <&fpga_bank0 15 0>;
+			debug-gpios = <&fpga_bank1 12 0>;
```
ts-kris: Replace the constant `0` in both `-gpios` with `GPIO_ACTIVE_HIGH`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+			interrupts = <21 IRQ_TYPE_LEVEL_HIGH>;
+		};
+	};
+
+	sfp0: sfp {
+		compatible = "sff,sfp";
+		i2c-bus = <&sfp0_i2c>;
+	};
+
+	sfp1: sfp {
```
ts-kris: These may need unique node names, e.g. `sfp0: sfp@0` since the node names are what the kernel uses to ID and the labels are intended more for human use

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+	buffer-manager = <&bm>;
+
+	fixed-link {
+		speed = <1000>;
+		full-duplex;
+	};
+};
+
+&gpio0 {
+	gpio-line-names =
```
ts-kris: Prefer a more compact format for `gpio-line-names` (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		"",
+		"MPP_27_GE_TXCLK",
+		"MPP_28_GE_TXD0",
+		"MPP_29_GE_TXD1",
+		"MPP_30_GE_TXD2",
+		"MPP_31_GE_TXD3";
+};
+
+&gpio1 {
+	gpio-line-names =
```
ts-kris: Prefer a more compact format for `gpio-line-names` (as many as possible on each line) to match other in-kernel FDTs.

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		"";
+};
+
+&i2c0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&i2c0_pins>;
+	clock-frequency = <100000>;
+	status = "okay";
+
+	silabs_wdt: silabs-wdt@54 {
```
ts-kris: Should be `silabs_wdt: watchdog@54`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		compatible = "technologic,ts7100-wdt";
+		reg = <0x54>;
+	};
+
+	rtc1: m41t00s@68 {
```
ts-kris: Should be `m41t00s: rtc@68`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+		compatible = "m41t00";
```
ts-kris: Should be `st,m41t00`

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+	clock-frequency = <100000>;
+	status = "okay";
+
+	silabs_wdt: silabs-wdt@54 {
+		compatible = "technologic,ts7100-wdt";
+		reg = <0x54>;
+	};
+
+	rtc1: m41t00s@68 {
+		compatible = "m41t00";
```
ts-kris: Should be `st,m41t00`

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,840 @@
...
+	pinctrl-names = "default";
+	status = "okay";
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_pins>;
+	status = "okay";
+	num-cs = <3>;
+	cs-gpios = <0>, <&fpga_bank1 9 0>;
```
ts-kris: Replace hardcoded `0` with `GPIO_ACTIVE_LOW` (LOW needed for SPI CS unless it really is active high, then it requires another property added)

arch/arm/configs/tsa38x_defconfig
ts-kris, line 1: I havn't yet had a chance to dig in to the two defconfigs added.

arch/arm/configs/tsa38x_minimal_defconfig
ts-kris, line 1: I havn't yet had a chance to dig in to the two defconfigs added.

ts-kris (changes requested):
There are also a few other outstanding things I found in testing that I need to look closer at.

I'm not sure if this needs to be hooked up to anything, otherwise pretty benign.
```
[    1.903152] hwmon hwmon0: temp1_input not attached to any thermal zone
```

I'm not sure if any of these should have supplies hooked up:
```
[    1.799195] ts7800v2-gpio e4081000.ts7800v2-gpio: ts7800v2_gpio driver probe successful.
[    1.810861] ahci-mvebu f10a8000.sata: supply ahci not found, using dummy regulator
[    1.818657] ahci-mvebu f10a8000.sata: supply phy not found, using dummy regulator
[    1.826252] ahci-mvebu f10a8000.sata: supply target not found, using dummy regulator
[    1.834182] ahci-mvebu f10a8000.sata: AHCI 0001.0000 32 slots 2 ports 6 Gbps 0x3 impl platform mode
[    1.843296] ahci-mvebu f10a8000.sata: flags: 64bit ncq sntf led only pmp fbs pio slum part sxs 
```

I also get this mess that is likely from WILC, but, since CAN is loading around the same time could be related to that and needs more investigation:
```
[   21.257241] wilc_spi: loading out-of-tree module taints kernel.
[   21.276664] WILC_SPI spi0.3: spiModalias: wilc3000, spiMax-Speed: 20000000
[   21.302257] Registering wifi device
[   21.302269] Max scan ids= 10,Max scan IE len= 1000,Signal Type= 1,Interface Modes= 844
[   21.302806] Initializing Locks ...
[   21.333443] WILC_SPI spi0.3: WILC SPI probe success
[   24.407205] CAN device driver interface
[   24.643061] sja1000 CAN netdevice driver
[   24.759641] sja1000_platform e408104c.can: sja1000_platform device registered (reg_base=4b410d77, irq=88)
[   25.391100] 8<--- cut here ---
[   25.394184] Unhandled fault: external abort on non-linefetch (0x1018) at 0xb6f73008
[   25.401873] pgd = b2e6298e
[   25.404593] [b6f73008] *pgd=03624831, *pte=fc081703, *ppte=fc081e33
[   25.444294] 8<--- cut here ---
[   25.447377] Unhandled fault: external abort on non-linefetch (0x1018) at 0xb6f94008
[   25.455066] pgd = b2e6298e
[   25.457786] [b6f94008] *pgd=03624831, *pte=fc081743, *ppte=fc081c33
[   25.751554] at_pwr_dev: open()
[   25.754716] AT PWR: bt_power_up
[   25.785014] WILC POWER UP
[   25.788265] at_pwr_dev: close()
[   25.792508] at_pwr_dev: open()
[   25.799586] at_pwr_dev: close()
```

ts-kris:
The following:
```
[   11.274062] EXT4-fs (tssdcarda1): re-mounted. Opts: (null)
[   11.556004] systemd-journald[1631]: Received client request to flush runtime journal.
[   22.980854] 8<--- cut here ---
[   22.983940] Unhandled fault: external abort on non-linefetch (0x1018) at 0xb6fa3008
[   22.991629] pgd = 904d4768
[   22.994350] [b6fa3008] *pgd=034d7831, *pte=fc081703, *ppte=fc081e33
[   23.005805] 8<--- cut here ---
[   23.008890] Unhandled fault: external abort on non-linefetch (0x1018) at 0xb6ffb008
[   23.016580] pgd = 707faaff
[   23.019300] [b6ffb008] *pgd=034d9831, *pte=fc081743, *ppte=fc081c33
root@ts-7800-v2:~# lsmod
Module                  Size  Used by
fuse                  114688  1
root@ts-7800-v2:~#
```

I can confirm was not caused by WILC or CAN. After blocking modules for both devices, this message still ends up appearing. I'll try to do a debug kernel build when I have a moment to see if I can get any more information out of that error.

EDIT: After updating to a newer rootfs from @luckshot, I no longer see this output. I'm not sure why or what was causing this.

EDIT: It looks like its caused by improper bus access from `ts7800ctl` which @luckshot did mention needed some changes for 5.10:
```
root@ts-7800-v2:~# ts7800ctl -i
model=7800
[ 3140.011113] 8<--- cut here ---
buildmodel=TS-78[ 3140.014289] Unhandled fault: external abort on non-linefetch (0x1018) at 0xb6f51000
00-V2-DMW3I
max[ 3140.023359] pgd = 4a7b5028
_rate=1333
max_[ 3140.027464] [b6f51000] *pgd=066ea831, *pte=b6f52703, *ppte=b6f52e33
cores=2
pcbrev=a
Bus error
```

arch/arm/boot/dts/armada-385-ts7800-v2.dts
```diff
@@ -0,0 +1,643 @@
...
+	};
+
+	led-controller {
+		compatible = "gpio-leds";
+
+		led-0 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_POWER;
+			gpios = <&ts7800v2_gpio 118 GPIO_ACTIVE_HIGH>;
+			linux,default-trigger = "heartbeat";
```
ts-kris: I just noticed this heartbeat trigger too, "I think this would be the only board we have green LED set up as a heartbeat. Not an issue either way, but, it begs the question, do we want to have this match other existing platforms or do we want other platforms matching this?"

This would put both a38x's so far as the only boards that do this.

arch/arm/boot/dts/armada-385-ts7840.dts
ts-kris: I just noticed the 7800v2 does this as well. It would make the a38x's the only boards that do this.

drivers/gpio/gpio-ts7800-v2.c
luckshot: I think this is a redundancy left over from a previous bug in u-boot.  It's probably safe to remove.

drivers/gpio/gpio-ts7820.c
markfeathers: Yes, this was some debug code that was left in.  Fixed in the latest version.

arch/arm/boot/dts/armada-385-ts7840.dts
markfeathers: Fixed in the latest

arch/arm/boot/dts/armada-385-ts7800-v2.dts
markfeathers: Fixed in latest

arch/arm/boot/dts/armada-385-ts7840.dts
markfeathers: Fixed in latest

arch/arm/boot/dts/armada-385-ts7800-v2.dts
markfeathers: This has been changed.  I still like it for an actual shipping application default, but it does make the image replicator difficult when blinking green is often used to show success.  Users can still enable it after, but with the latest its no longer the default.

drivers/mmc/host/tssdcard.c
ts-kris, line 1: Probably not worth it to get this driver merged, but, we do a lot of manual struct bio management by hand here. The bio API (which we already use a number of functions of) offers list control functions so we don't need to deal with that explicitly.

arch/arm/boot/dts/armada-385-ts7840.dts
```diff
@@ -0,0 +1,731 @@
...
+		"CPU_TYPE_0", "", "", "", "", "", "GPS_PPS", "ETH_SW_INT#",
+		"CPU_TYPE_1", "", "", "", "", "FPGA_FLASH_SELECT", "EMMC_CMD",
+		"SPREAD_SPECTRUM#", "", "DETECT_9478", "", "", "", "", "", "",
+		"", "", "", "";
+};
+
+&i2c0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&i2c0_pins>;
+	clock-frequency = <100000>;
```
ts-kris, line 552: I pointed this out on the 7800v2 DTS but missed it here. Is there any reason this bus is limited to 100 kHz?

ts-kris (commented):
A couple of unresolved comments and some new ones added. Please review/respond/resolve.

I've been doing broad, high-level testing of the unit on my desk with these changes and everything looks good so far. Once remaining comments are resolved I'll approve.

Links to outstanding comments for easier reference:
~~https://github.com/embeddedTS/linux-lts/pull/18#discussion_r1152574101~~
~~https://github.com/embeddedTS/linux-lts/pull/18#discussion_r1152593238~~
~~https://github.com/embeddedTS/linux-lts/pull/18#discussion_r1152595969~~

arch/arm/boot/dts/armada-385-ts7800-v2.dts
markfeathers: Looks like the RTC at least is the limit: Serial interface supports I2C bus (100 kHz protocol)

markfeathers: Wrong about the rtc, that was the M41T00 (not S), S is 400khz, but the microcontroller may not be set up for 400khz.  Leaving at 100khz for now.

arch/arm/boot/dts/armada-385-ts7840.dts
markfeathers: Leaving at 100khz since the microcontroller may not support 400khz.

arch/arm/boot/dts/armada-385-ts7800-v2.dts
markfeathers: Just pushed a fix for this.

drivers/watchdog/ts7100_wdt.c
markfeathers: Just pushed a fix for this.

ts-kris (approved):
All looks good to me. All original change requests were addressed in one way or another.

markfeathers (approved):

PR #20, opened 2023-03-21
mfd: tssupervisor-core: Add serial support

markfeathers:
Added support for a "serial" sysfs entry that can be written once in the product's lifetime

### DIFF
--- a/drivers/mfd/tssupervisor-core.c
+++ b/drivers/mfd/tssupervisor-core.c
@@ -33,12 +33,14 @@ static const struct regmap_range ts_supervisor_read_regs[] = {
 	regmap_reg_range(16, 16), /* flags */
 	regmap_reg_range(24, 24), /* inputs */
 	regmap_reg_range(32, 32), /* reboot_reason */
+	regmap_reg_range(34, 37), /* serial */
 	regmap_reg_range(128, 160), /* ADCs+temp */
 };
 
 static const struct regmap_range ts_supervisor_write_regs[] = {
 	regmap_reg_range(8, 8), /* cmds */
 	regmap_reg_range(16, 16), /* flags */
+	regmap_reg_range(34, 37), /* serial */
 };
 
 const struct regmap_access_table ts_supervisor_read_register_set = {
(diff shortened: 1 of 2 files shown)


drivers/mfd/tssupervisor-core.c
```diff
@@ -119,6 +121,66 @@ static ssize_t wake_en_show(struct device *dev,
...
+		return ret;
+
+	/* Dont write the SN if its already written */
+	if (ctrl & SUPER_SN_LOCKED)
+		return -EEXIST;
+
+	if (count < 17)
+		return -EIO;
+
+	if (sscanf(buf, "%02hhx:%02hhx:%02hhx:%02hhx:%02hhx:%02hhx",
```
ts-kris, line 143: Is length of buf guaranteed? If not, we should probably do a strlen check or something before trying to parse it.

drivers/mfd/tssupervisor-core.c
```diff
@@ -119,6 +121,66 @@ static ssize_t wake_en_show(struct device *dev,
...
+		return ret;
+
+	/* Dont show the serial number if its not written yet */
+	if (!(ctrl & SUPER_SN_LOCKED))
+		return -EINVAL;
+
+	ret = regmap_bulk_read(super->regmap, SUPER_SERIAL0, sn, 3);
+	if (ret)
+		return ret;
+	ret = sprintf(buf, "%02x:%02x:%02x:%02x:%02x:%02x\n",
```
ts-kris, line 185: Same with above. Is buf guaranteed to be any specific size? We may want to check its length before writing to it if not. Could also use snprintf on buf, but I don't think thats hugely necessary if we do length check/are guaranteed buf size.

drivers/mfd/tssupervisor-core.c
```diff
@@ -119,6 +121,66 @@ static ssize_t wake_en_show(struct device *dev,
...
+
+	if (sscanf(buf, "%02hhx:%02hhx:%02hhx:%02hhx:%02hhx:%02hhx",
+		   &sn[0], &sn[1], &sn[2], &sn[3], &sn[4], &sn[5]) != 6)
+		return -EINVAL;
+
+	ret = regmap_bulk_write(super->regmap, SUPER_SERIAL0, sn, 3);
+	if (ret)
+		return ret;
+
+	ret = regmap_update_bits(super->regmap, SUPER_SERIAL_CTRL,
```
ts-kris, line 159: Maybe add a readback check to ensure the data we expect to be written was written before locking out future writes?

drivers/mfd/tssupervisor-core.c
markfeathers: The length of buf is specified by count. I am making sure count is at least the length I expect for a mac address before attempting to parse buf, but I did not implement an upper bound so we don't have to deal with differences of echo or an application that may or may not add "\n".

luckshot: I think that's what he's doing here, isn't it?  If sscanf() doesn't return 6 characters exactly, return -einval?

drivers/mfd/tssupervisor-core.c
markfeathers: The actual locking out of future writes happens on the microcontroller if the write succeeds.  If it were to fail to write for some reason then the SN_LOCKED bit would not be set the next time its read.  If you mean to double check the i2c write, we should fail out before them if the bulk_write fails before locking out the mac address.

drivers/mfd/tssupervisor-core.c
ts-kris: Oops, I had the src and dest backwards here. So long as the incoming string is formatted for sscanf to be happy, there arn't any issues here.

drivers/mfd/tssupervisor-core.c
markfeathers: From:
https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt

```
sysfs allocates a buffer of size (PAGE_SIZE) and passes it to the
method. Sysfs will call the method exactly once for each read or
write. This forces the following behavior on the method
implementations: 
```
IIRC PAGE_SIZE should be 4K, certainly larger than the 18 bytes we need.

drivers/mfd/tssupervisor-core.c
ts-kris: So, if there is a bit flip on I2C, we have no way to read back what we tried to write and check that the written SN is what was intended to be written and correct if there was an issue?

drivers/mfd/tssupervisor-core.c
ts-kris: Ah! I wasn't aware that was how the buffer was allocated and handled.

luckshot (dismissed):

drivers/mfd/tssupervisor-core.c
markfeathers: Just force pushed a change for this to check the OTP data before writing it.

ts-kris (approved):

luckshot (approved):
Goku:
You can't be serial!   
Vegeta:
Oh... I'm SUPER SERIAL.

PR #24, opened 2023-04-18
Config file cleanup for 5.10.y

ts-kris:
Misc defconfig changes and tweaks we discovered in further testing.


ts-kris (commented):
Is this actually necessary to have as a builtin? IIRC, these should load automatically if they are correctly defined in the FDT. Was there a situation where the modules weren't loading?

I'm operating and questioning this on the assumption that, things that are critical to operation of a device or its boot speed, should be builtin; everything else can usually be okay being a module. Since MTD isn't needed for boot, nor early operation of the device, I'd think so long as these are being correctly loaded they could stay as modules.

arch/arm/configs/tsimx6ul_minimal_defconfig
```diff
@@ -78,16 +78,16 @@ CONFIG_FW_LOADER_USER_HELPER_FALLBACK=y
...
-CONFIG_MTD_BLOCK=m
+CONFIG_MTD_BLOCK=y
 CONFIG_MTD_CFI=m
 CONFIG_MTD_JEDECPROBE=m
 CONFIG_MTD_CFI_INTELEXT=m
 CONFIG_MTD_CFI_AMDSTD=m
 CONFIG_MTD_CFI_STAA=m
 CONFIG_MTD_SST25L=m
-CONFIG_MTD_SPI_NOR=m
+CONFIG_MTD_SPI_NOR=y
```
ts-kris, line 89: Is this actually necessary to have as a builtin? IIRC, these should load automatically if they are correctly defined in the FDT. Was there a situation where the modules weren't loading?

I'm operating and questioning this on the assumption that, things that are critical to operation of a device or its boot speed, should be builtin; everything else can usually be okay being a module. Since MTD isn't needed for boot, nor early operation of the device, I'd think so long as these are being correctly loaded they could stay as modules.

ts-kris:
Everything looks good to me. Ive added setting MTD to builtin on imx6 platforms to match other defconfigs. I also squashed the both imx28 full and minimal WDT auto-feed changes together.

markfeathers (dismissed):

lionel:
These should be getting built/installed, no?
```
root@tsimx6:~# modprobe st_magn
modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/5.10.170/modules.dep.bin'
modprobe: FATAL: Module st_magn not found in directory /lib/modules/5.10.170
root@tsimx6:~# modprobe pps_gpio
modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/5.10.170/modules.dep.bin'
modprobe: FATAL: Module pps_gpio not found in directory /lib/modules/5.10.170
root@tsimx6:~# modprobe st_sensors
modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/5.10.170/modules.dep.bin'
modprobe: FATAL: Module st_sensors not found in directory /lib/modules/5.10.170
root@tsimx6:~# 
```

ts-kris:
What platform are you testing this on? I'm not aware of `pps-gpio` on any i.MX6 board being readily used. And I can't remember which platforms might have accelerometer/magnetometer builtin.

FWIW `tsimx6_defconfig` builds the `st_...` modules for accelerometer/magnetometer. Otherwise both build `mma8452` in to the kernel.

EDIT: Ah! TS-TPC-8390-4900 uses `pps-gpio`! I'll add that to both configs as a module. Let me know about the accelerometer/magnetometer as it looks like every imx6 platform uses the `mma8452`

lionel:
Perhaps mx6 boards use only that Freescale part, but all the latest (non-Invensese) 6ul boards I'm aware of use a pair of ST Micro parts and require (in 5.10):
```
CONFIG_ST_ISM330DLC_IIO=m
CONFIG_ST_MAG40_IIO=m
```

lionel:
Also, besides the gyro/mag issue in both 6ul defconfigs, the minimal one has no IIO triggers at all, and `tsimx6ul_defconfig` is missing:
```
CONFIG_IIO_HRTIMER_TRIGGER=m
CONFIG_IIO_TIGHTLOOP_TRIGGER=m
```

lionel:
Also, in documenting the defconfigs, it may be worth noting what is required for our all of our documentation to be valid, as I assume that the minimal defconfig is not expected to work with all documentation examples? Or is it ?-)

ts-kris:
Sounds good. When I get back in the office next week I'll get these added in. I was not aware there was a change as I'm only aware of Invensense parts being used.

> Perhaps mx6 boards use only that Freescale part, but all the latest (non-Invensese) 6ul boards I'm aware of...

Your example is a booted i.MX6 (based on the hostname) which is why I went there specifically. Should the ST parts be added in all of i.MX6 and i.MX6UL or just i.MX6?

> Also, in documenting the defconfigs, it may be worth noting what is required for our all of our documentation to be valid, as I assume that the minimal defconfig is not expected to work with all documentation examples? Or is it ?-)

The minimal defconfig is expected to:
- Support every COTS hardware configuration of the platform (does not include any customized devices).
- Support some common USB devices, UMS, FTDI, etc.
- Support all of the documented examples in the platform manual.

The non-minimal defconfig for each platform is meant to go above and beyond that, adding in support for more USB devices, more esoteric networking options, etc.

The distinction between a module and built-in for these things depends on size and urgency and is pretty subjective. 

So, if there is any support still missing that you are aware of, please let me know what config options are needed for which platforms and I can get those set up. You are also welcome to push said changes to this branch and I can review, rebase/squash as needed, and go from there.

lionel:
> > Perhaps mx6 boards use only that Freescale part, but all the latest (non-Invensese) 6ul boards I'm aware of...
> 
> Your example is a booted i.MX6 (based on the hostname) which is why I went there specifically. Should the ST parts be added in all of i.MX6 and i.MX6UL or just i.MX6?

At the moment, the only boards using the ST Micro parts are 6ULs.

> So, if there is any support still missing that you are aware of, please let me know what config options are needed for which platforms and I can get those set up. You are also welcome to push said changes to this branch and I can review, rebase/squash as needed, and go from there.

The 7100 and 7180 are still using `CONFIG_TS7100_WATCHDOG`.

ts-kris:
@lionel Can you confirm these config names?
```
CONFIG_ST_ISM330DLC_IIO=m
CONFIG_ST_MAG40_IIO=m
```
They do not appear to be valid when searching for them. Did you mean?
```
CONFIG_IIO_ST_LSM6DSX
```
(Claims support for ism330dlc and is already in both `tsimx6ul_defconfig` and `tsimx6ul_minimal_defconfig`)


I can't find any documentation or details matching "mag40" as a name, the closest I could find is:
```
CONFIG_IIO_ST_MAGN_3AXIS
```
Which is already in both `tsimx6ul_defconfig` and `tsimxtul_minimal_defconfig`

lionel:
When not everything showed up, my brain went all 4.9 on me. Oops! Let me see a little more closely what I did to make it work on the 7180.

lionel (dismissed):
I notice most minimal defconfigs still are missing PPS (except for the imx28), which you mentioned is used with one of the baseboards, perhaps?  Also I've noticed that PTP support isn't compiled into any minimal defconfig. But otherwise all looks fine for now. I can add those later to the 6ul_minimal_defconfig with the TS-71x0 boards.

markfeathers (approved):

randytsi369 (approved):

PR #25, opened 2023-07-24
ARM: dts: imx6: TS-4900 set phy skew

markfeathers:
Set data to minimum, and clock to maximum. Without this the phy does not link at 1000M on the solo core.


ts-kris (approved):
Looks good to me. Changes not tested, review only.

randytsi369 (approved):

PR #26, opened 2023-07-24
ARM: configs: tsimx6ul: Add CONFIG_I2C_OCORES=m

markfeathers:
Needed to support the mikrobus interface on the TS-7250-V3.


ts-kris (approved):
Looks good to me, touches both minimal and regular defconfig. Review only, untested.

lionel (approved):

PR #27, opened 2023-08-01
serial: 8250: 8250_ts: use mapbase for base addr

markfeathers:
This fixes an issue with mapbase/membase/iobase not being used. If none of those are set, the driver loads and appears to work but never calls uart_config on startup. It then ignores the value that should be initialized for FCR. This causes the fifo to be disabled.

This driver also now handles the serial in/out hooks in this 8250_ts.c driver.


ts-kris:
Does this PR get our fingers completely out of upstream 8250 source files then?

randytsi369 (approved):
Changes reviewed - Not tested.

markfeathers:
>  Does this PR get our fingers completely out of upstream 8250 source files then?

Yes :)

drivers/tty/serial/8250/8250_port.c
```diff
@@ -459,25 +458,6 @@ static void io_serial_out(struct uart_port *p, int offset, int value)
 	outb(value, p->iobase + offset);
 }
 
-#ifdef CONFIG_SERIAL_8250_TS
```
ts-kris, line 462: Is this CONFIG_ option still otherwise used? If not, that should also be removed from Kconfig files.

drivers/tty/serial/8250/8250_ts.c
```diff
@@ -7,51 +7,52 @@
...
 	port->flags = UPF_SHARE_IRQ | UPF_BOOT_AUTOCONF |
-				UPF_FIXED_PORT | UPF_FIXED_TYPE;
-	port->iotype = UPIO_TSISABUS;
+		      UPF_FIXED_PORT | UPF_FIXED_TYPE;
+	port->iotype = UPIO_MEM;
 	port->irqflags = IRQF_TRIGGER_HIGH;
-	port->private_data = priv;
+	port->private_data = platform_get_drvdata(to_platform_device(pdev->dev.parent));
 	port->uartclk = 1843200;
-	port->fifosize = 16;
```
ts-kris, line 44: I'm guessing fifosize is inferred by port->type?

markfeathers: Correct. I tested this out with a printk to make sure it was 16 bytes during normal operation after removing this line.

ts-kris (approved):
Sounds good to me then. Changed not tested, only reviewed.

PR #41, opened 2023-09-25
spidev warning fix

markfeathers:


ts-kris:
I know this is only a draft but I'm immediately going to comment. Please use `technologic,spi-header` to match our existing in-kernel prefix name: https://github.com/embeddedTS/linux-lts/blob/linux-5.10.y/Documentation/devicetree/bindings/vendor-prefixes.yaml#L1058-L1059

We decided to not to approach changing this name upstream, yet, and we can certainly consider it again. However, keep it `technologic` for the time being so when we push changes upstream and mirror them here, we don't have to hunt down multiple references.

ts-kris (changes requested):
Fix the commit title of cc146a2 and you've otherwise got my approval.

markfeathers:
That should be fixed now

randytsi369 (approved):

ts-kris (approved):
Looks good to me. Not tested.

PR #44, opened 2023-10-05
Kernel config fixup

markfeathers:
Add config fixes found from CI testing.


luckshot (dismissed):

arch/arm/configs/tsimx6_defconfig
```diff
@@ -28,8 +28,8 @@ CONFIG_NAMESPACES=y
 CONFIG_USER_NS=y
 CONFIG_CHECKPOINT_RESTORE=y
 CONFIG_RELAY=y
-CONFIG_BLK_DEV_INITRD=y
```
ts-kris, line 31: I'm 95% sure this is needed to boot from ramdisk, e.g. Image Replicator loading CPIO in to RAM and booting to that (granted IR tools for most platforms are still running 4.9 at the moment, of which those configs all have this option enabled). I would not recommend removing this without testing to confirm.

But also, this option is still in tsimx6ul_defconfig, so if it does need to be removed, it should be removed from there too.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -521,17 +532,52 @@ CONFIG_SATA_AHCI_PLATFORM=y
...
+CONFIG_DM_RAID=m
+CONFIG_DM_ZERO=m
+CONFIG_DM_MULTIPATH=m
+CONFIG_DM_DELAY=m
+CONFIG_DM_DUST=m
+CONFIG_DM_FLAKEY=m
+CONFIG_DM_VERITY=m
+CONFIG_DM_SWITCH=m
+CONFIG_DM_LOG_WRITES=m
+CONFIG_DM_INTEGRITY=m
```
ts-kris: I would think if we're enabling these on one defconfig, we should probably enable on all? Things like the about for device mapper support could easily apply to other platforms. Personally I think that is too much kitchen sink, and various MD device support along with the more esoteric device mapper options should be left up to the user. But I'm more concerned with uniform support if you feel those options are even remotely useful.

Note: My comment is not limited to just MD/DM options highlighted here, there are a handful of other options enabled in this defconfig that are not present in any other current defconfig.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -746,8 +820,27 @@ CONFIG_REGULATOR_USERSPACE_CONSUMER=m
...
+CONFIG_IR_IMON_DECODER=m
+CONFIG_IR_RCMM_DECODER=m
 CONFIG_RC_DEVICES=y
+CONFIG_RC_ATI_REMOTE=m
+CONFIG_IR_IGORPLUGUSB=m
+CONFIG_IR_IGUANA=m
+CONFIG_IR_TTUSBIR=m
 CONFIG_IR_GPIO_CIR=m
+CONFIG_IR_GPIO_TX=m
+CONFIG_IR_PWM_TX=m
```
ts-kris: These platforms don't have any IR hardware? Anything that would be USB would take care of the decoding on the USB dongle.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -1053,15 +1267,21 @@ CONFIG_EXFAT_FS=m
...
 CONFIG_NTFS_RW=y
 CONFIG_TMPFS_POSIX_ACL=y
+CONFIG_HFS_FS=m
 CONFIG_CRAMFS=m
 CONFIG_SQUASHFS=m
+CONFIG_SQUASHFS_XATTR=y
 CONFIG_SQUASHFS_LZ4=y
 CONFIG_SQUASHFS_LZO=y
 CONFIG_SQUASHFS_XZ=y
+CONFIG_SQUASHFS_ZSTD=y
```
ts-kris: SquashFS doesn't look like its enabled anywhere else. I don't think this is useful as a kitchen sink.

ts-kris (changes requested):
I'd like to see 1 of 2 things. The tsimx6_defconfig change adds a LOT of additional support.
1) I think most of these are unneeded or not useful, even as a kitchen sink. I commented on a couple, but I did not go through all of them. Many of them look like more advanced/esoteric options that I do not think would ever be useful for customers. I think the goal of our kitchen sinking should target out of the box support coverage for, say, 75% of customers, rather than trying to include every possible cornet case.

2) If some of the options really are necessary or even just useful, then many of them likely need to also be enabled in the tsimx6ul_ and tsa38x_ defconfigs.

luckshot (dismissed):

arch/arm/configs/tsimx6_defconfig
```diff
@@ -1,88 +1,80 @@
...
-CONFIG_RELAY=y
 CONFIG_BLK_DEV_INITRD=y
-# CONFIG_RD_ZSTD is not set
-CONFIG_EXPERT=y
-CONFIG_KALLSYMS_ALL=y
+CONFIG_CC_OPTIMIZE_FOR_SIZE=y
 CONFIG_BPF_SYSCALL=y
-# CONFIG_SLUB_DEBUG is not set
+# CONFIG_RSEQ is not set
+CONFIG_EMBEDDED=y
```
ts-kris: Currently, this is only enabled for 6 and 6ul minimal. This doesn't appear to have been added to tsa38x or tsimx28 configs. Any particular reason for inclusion vs not including this option?

arch/arm/configs/tsimx6_defconfig
```diff
@@ -1,88 +1,80 @@
...
+# CONFIG_VM_EVENT_COUNTERS is not set
 # CONFIG_COMPAT_BRK is not set
-CONFIG_PROFILING=y
+CONFIG_SLOB=y
```
ts-kris: as EMBEDDED above, this is only enabled right now for 6 and 6ul minimal, and is only added to 6 defconfig here. Any reason why this would be included vs not?

arch/arm/configs/tsimx6_defconfig
```diff
@@ -1,88 +1,80 @@
...
+# CONFIG_SLAB_MERGE_DEFAULT is not set
+CONFIG_ARCH_MULTI_V6=y
 CONFIG_ARCH_MXC=y
 CONFIG_SOC_IMX6Q=y
 CONFIG_ARM_ERRATA_814220=y
 CONFIG_SMP=y
 CONFIG_HAVE_ARM_ARCH_TIMER=y
+CONFIG_HOTPLUG_CPU=y
```
ts-kris, line 35: I really don't think we need hotplug CPU

arch/arm/configs/tsimx6_defconfig
```diff
@@ -1,88 +1,80 @@
...
 CONFIG_ARM_PSCI=y
-CONFIG_HIGHMEM=y
```
ts-kris, line 45: Why is this removed here and not anywhere else? It also seems like we probably don't need this anywhere on our current models. According to config help, removing this would result in a slightly faster kernel. 

tl;dr this option is if there is more than 4 GB total of mem/io space.

arch/arm/configs/tsa38x_defconfig
```diff
@@ -495,239 +400,228 @@ CONFIG_CAN_GS_USB=m
...
-CONFIG_CONNECTOR=m
 CONFIG_MTD=y
 CONFIG_MTD_CMDLINE_PARTS=y
 CONFIG_MTD_BLOCK=y
 CONFIG_MTD_CFI=y
 CONFIG_MTD_CFI_INTELEXT=y
 CONFIG_MTD_CFI_AMDSTD=y
 CONFIG_MTD_CFI_STAA=y
+CONFIG_MTD_RAW_NAND=y
+CONFIG_MTD_NAND_MARVELL=y
```
ts-kris: This is added to this config but removed from others. None of our devices have NAND anymore. Is this really needed?

arch/arm/configs/tsimx6_defconfig
```diff
@@ -712,193 +704,58 @@ CONFIG_I2C_ALGOPCA=m
...
 CONFIG_DRM=y
 CONFIG_DRM_PANEL_LVDS=y
 CONFIG_DRM_PANEL_SIMPLE=y
+CONFIG_DRM_PANEL_SEIKO_43WVF1G=y
 CONFIG_DRM_TI_TFP410=y
 CONFIG_DRM_DW_HDMI_AHB_AUDIO=m
 CONFIG_DRM_DW_HDMI_CEC=y
 CONFIG_DRM_IMX=y
 CONFIG_DRM_IMX_PARALLEL_DISPLAY=y
+CONFIG_DRM_IMX_TVE=y
```
ts-kris: This is for supporting TV Encoders on the imx53, we don't need this.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -907,56 +764,79 @@ CONFIG_SND_SOC_IMX_ES8328=m
...
-CONFIG_USB_CONFIGFS_F_UVC=y
-CONFIG_USB_CONFIGFS_F_PRINTER=y
-CONFIG_USB_ZERO=m
-CONFIG_USB_AUDIO=m
+CONFIG_USB_CONFIGFS=m
 CONFIG_USB_ETH=m
 CONFIG_USB_G_NCM=m
 CONFIG_USB_GADGETFS=m
-CONFIG_USB_FUNCTIONFS=m
-CONFIG_USB_MASS_STORAGE=m
```
ts-kris, line 959: Do we want to remove OTG mass storage support?

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -933,51 +783,10 @@ CONFIG_USB_SERIAL_WISHBONE=m
...
-CONFIG_USB_ZERO=m
-CONFIG_USB_AUDIO=m
-CONFIG_USB_ETH=m
-CONFIG_USB_G_NCM=m
-CONFIG_USB_GADGETFS=m
-CONFIG_USB_FUNCTIONFS=m
-CONFIG_USB_FUNCTIONFS_ETH=y
-CONFIG_USB_FUNCTIONFS_RNDIS=y
-CONFIG_USB_FUNCTIONFS_GENERIC=y
-CONFIG_USB_MASS_STORAGE=m
```
ts-kris, line 970: Do we want to remove OTG mass storage support?

arch/arm/configs/tsa38x_defconfig
```diff
@@ -980,113 +772,65 @@ CONFIG_MMC_MVSDIO=y
...
-CONFIG_LEDS_TRIGGER_BACKLIGHT=m
-CONFIG_LEDS_TRIGGER_CPU=y
-CONFIG_LEDS_TRIGGER_ACTIVITY=y
-CONFIG_LEDS_TRIGGER_GPIO=m
-CONFIG_LEDS_TRIGGER_DEFAULT_ON=m
-CONFIG_LEDS_TRIGGER_TRANSIENT=m
-CONFIG_LEDS_TRIGGER_CAMERA=m
-CONFIG_LEDS_TRIGGER_PANIC=y
-CONFIG_LEDS_TRIGGER_NETDEV=y
-CONFIG_LEDS_TRIGGER_PATTERN=y
```
ts-kris: Why are all of these LED triggers being removed, multiple configs, not just this one. Things like netdev, hearbeat, cpu/activity, panic, disk, etc. might be useful for some customers and we've had them supported for a while now.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -452,31 +416,36 @@ CONFIG_BT_RFCOMM_TTY=y
...
 CONFIG_BT_HCIUART_QCA=y
+CONFIG_BT_HCIUART_AG6XX=y
+CONFIG_BT_HCIUART_MRVL=y
+CONFIG_BT_HCIBCM203X=m
+CONFIG_BT_HCIBPA10X=m
+CONFIG_BT_HCIBFUSB=m
+CONFIG_BT_HCIVHCI=m
+CONFIG_BT_MRVL=m
+CONFIG_BT_ATH3K=m
+CONFIG_BT_MTKUART=m
```
ts-kris, line 439: I don't think we need all of the UART support. USB for sure, and UART support for any modules that might be soldered down is reasonable IMO. This is in multiple configs I believe.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -524,50 +482,103 @@ CONFIG_MD=y
...
+# CONFIG_NET_VENDOR_ALTEON is not set
+# CONFIG_NET_VENDOR_AMAZON is not set
+# CONFIG_NET_VENDOR_AMD is not set
+# CONFIG_NET_VENDOR_AQUANTIA is not set
+# CONFIG_NET_VENDOR_ARC is not set
+# CONFIG_NET_VENDOR_ATHEROS is not set
+# CONFIG_NET_VENDOR_AURORA is not set
 # CONFIG_NET_VENDOR_BROADCOM is not set
-CONFIG_CS89x0=y
-CONFIG_CS89x0_PLATFORM=y
```
ts-kris, line 537: Very unlikely needed, but, this is the module used in the TS-ETH2. If we expect supporting that module on any platform, we should keep that in here. That includes 6ul for platforms that have PC/104 support

arch/arm/configs/tsimx6_defconfig
```diff
@@ -712,193 +704,58 @@ CONFIG_I2C_ALGOPCA=m
...
-CONFIG_GPIO_PCA953X=y
-CONFIG_GPIO_PCF857X=y
 CONFIG_GPIO_TS4900=y
-CONFIG_GPIO_74X164=y
 CONFIG_POWER_RESET=y
 CONFIG_POWER_RESET_SYSCON=y
 CONFIG_POWER_RESET_SYSCON_POWEROFF=y
-CONFIG_POWER_SUPPLY=y
-CONFIG_SENSORS_GPIO_FAN=y
-CONFIG_SENSORS_IIO_HWMON=m
```
ts-kris, line 736: I thought we needed this? This might also be selected by another driver and therefore is implied in the defconfig though.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -647,61 +673,27 @@ CONFIG_RSI_91X=m
...
 CONFIG_QTNFMAC_PCIE=m
-CONFIG_USB_NET_RNDIS_WLAN=m
-CONFIG_VIRT_WIFI=m
+# CONFIG_INPUT_LEDS is not set
 CONFIG_INPUT_MOUSEDEV=m
-CONFIG_INPUT_JOYDEV=m
 CONFIG_INPUT_EVDEV=y
+# CONFIG_KEYBOARD_ATKBD is not set
 CONFIG_KEYBOARD_GPIO=y
-CONFIG_KEYBOARD_SNVS_PWRKEY=y
```
ts-kris, line 658: I don't think this is an issue on imx6, but, be careful removing this as this changes the input event numbering. I know this is an issue in the imx6ul, specifically ts7553v2 for the keypad demos. But I don't believe this is an issue to remove here.

arch/arm/configs/tsimx6_defconfig
```diff
@@ -965,102 +845,59 @@ CONFIG_MMC_SDHCI_ESDHC_IMX=y
...
-CONFIG_IIO_SW_TRIGGER=m
-CONFIG_IIO_TRIGGERED_EVENT=m
-CONFIG_IIO_ST_ACCEL_3AXIS=m
-CONFIG_MMA8452=m
-CONFIG_IMX7D_ADC=m
-CONFIG_MCP320X=m
-CONFIG_VF610_ADC=m
-CONFIG_IIO_RESCALE=m
+CONFIG_IIO=y
+CONFIG_MMA8452=y
```
ts-kris, line 888: If this is optional on any builds, this should probably be left as a module? Reduce bloat on platforms that don't have it.

There are actually a handful of sensors that moved from modules to builtins. I know these have also changed over time, so I think these should all be modules unless there is reason to keep them as builtins.

This applies to multiple configs too it looks like.

arch/arm/configs/tsimx6_minimal_defconfig
```diff
@@ -235,7 +236,6 @@ CONFIG_GPIO_TS4900=y
 CONFIG_POWER_RESET=y
 CONFIG_POWER_RESET_SYSCON=y
 CONFIG_POWER_RESET_SYSCON_POWEROFF=y
-# CONFIG_HWMON is not set
```
ts-kris, line 238: Unless this was removed because another option selected this, we should keep this in for temperature sensor in CPU.

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -469,76 +601,73 @@ CONFIG_USB_VL600=m
...
+CONFIG_SERIAL_8250=m
 CONFIG_SERIAL_8250_NR_UARTS=24
 CONFIG_SERIAL_8250_RUNTIME_UARTS=24
-CONFIG_SERIAL_8250_TS=m
-CONFIG_SERIAL_OF_PLATFORM=y
+CONFIG_SERIAL_OF_PLATFORM=m
 CONFIG_SERIAL_IMX=y
 CONFIG_SERIAL_IMX_CONSOLE=y
+CONFIG_SERIAL_FSL_LPUART=y
+CONFIG_SERIAL_FSL_LPUART_CONSOLE=y
```
ts-kris: There are not LPUARTs on this platform.

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -857,26 +714,19 @@ CONFIG_DRM_IMX=y
 CONFIG_DRM_IMX_PARALLEL_DISPLAY=y
 CONFIG_DRM_MXSFB=y
 CONFIG_FB_MODE_HELPERS=y
+# CONFIG_FB_MX3 is not set
 CONFIG_FB_ST7565P=m
-CONFIG_DRM_FBDEV_EMULATION=y
```
ts-kris, line 861: We definitely need this. This actually should probably be added to all configs, but for sure would be needed on 6ul with display support.

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -857,26 +714,19 @@ CONFIG_DRM_IMX=y
...
 CONFIG_LCD_CLASS_DEVICE=y
+CONFIG_LCD_PLATFORM=y
 CONFIG_BACKLIGHT_CLASS_DEVICE=y
-CONFIG_BACKLIGHT_PWM=m
-CONFIG_BACKLIGHT_GPIO=m
-CONFIG_SOUND=y
-CONFIG_SND=y
-# CONFIG_SND_ARM is not set
-# CONFIG_SND_SPI is not set
-CONFIG_SND_USB_AUDIO=m
```
ts-kris, line 870: USB audio support not removed from every platform? I think we should keep this in as a module just in case. But if not, should be removed from everything.

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -857,26 +714,19 @@ CONFIG_DRM_IMX=y
...
-CONFIG_HID_MULTITOUCH=y
+CONFIG_BACKLIGHT_PWM=y
+CONFIG_BACKLIGHT_GPIO=y
+CONFIG_HID_MULTITOUCH=m
+CONFIG_USB_HID=m
 CONFIG_USB=y
 CONFIG_USB_ANNOUNCE_NEW_DEVICES=y
-CONFIG_USB_OTG=y
-CONFIG_USB_OTG_FSM=m
-CONFIG_USB_LEDS_TRIGGER_USBPORT=m
```
ts-kris: 6ul absolutely supports OTG, keep this in.

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -986,46 +795,23 @@ CONFIG_NEW_LEDS=y
...
-CONFIG_LEDS_TRIGGER_PANIC=y
-CONFIG_LEDS_TRIGGER_NETDEV=m
-CONFIG_LEDS_TRIGGER_PATTERN=m
-CONFIG_LEDS_TRIGGER_AUDIO=m
 CONFIG_RTC_CLASS=y
+# CONFIG_RTC_NVMEM is not set
 CONFIG_RTC_INTF_DEV_UIE_EMUL=y
 CONFIG_RTC_DRV_DS1307=y
 CONFIG_RTC_DRV_TSSUPERVISOR=y
-CONFIG_RTC_DRV_MXC=y
```
ts-kris, line 1008: This needs to be kept in as we use CPU RTC on 4100 standalone operation.

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -1095,12 +873,12 @@ CONFIG_CRC_T10DIF=y
 CONFIG_CRC7=m
 CONFIG_DMA_CMA=y
 CONFIG_CMA_SIZE_MBYTES=64
-CONFIG_PRINTK_TIME=y
```
ts-kris, line 1098: Don't we want to leave time in on the full defconfigs?

arch/arm/configs/tsimx6_defconfig
```diff
@@ -1069,27 +906,27 @@ CONFIG_NLS_ASCII=y
...
 CONFIG_CRYPTO_DEV_FSL_CAAM=y
+CONFIG_CRYPTO_DEV_SAHARA=y
+CONFIG_CRYPTO_DEV_MXS_DCP=y
 CONFIG_CRC_T10DIF=y
 CONFIG_CRC7=m
 CONFIG_CMA_SIZE_MBYTES=64
 CONFIG_FONTS=y
 CONFIG_FONT_8x8=y
 CONFIG_FONT_8x16=y
-CONFIG_PRINTK_TIME=y
```
ts-kris, line 1087: Don't we want to leave printk time in on full defconfigs?

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -1095,12 +873,12 @@ CONFIG_CRC_T10DIF=y
...
+# CONFIG_SYMBOLIC_ERRNAME is not set
 # CONFIG_DEBUG_BUGVERBOSE is not set
-CONFIG_DEBUG_FS=y
+# CONFIG_ENABLE_MUST_CHECK is not set
 # CONFIG_DEBUG_MISC is not set
 # CONFIG_SCHED_DEBUG is not set
 # CONFIG_DEBUG_PREEMPT is not set
-CONFIG_STACKTRACE=y
```
ts-kris, line 1104: Don't we want to leave stacktrace in too? In multiple configs

arch/arm/configs/tsimx6ul_defconfig
```diff
@@ -16,81 +14,134 @@ CONFIG_CFS_BANDWIDTH=y
...
 CONFIG_VFP=y
 CONFIG_NEON=y
+# CONFIG_SUSPEND is not set
+CONFIG_PM=y
 CONFIG_PM_DEBUG=y
-CONFIG_PM_TEST_SUSPEND=y
+# CONFIG_STACKPROTECTOR is not set
 CONFIG_MODULES=y
 CONFIG_MODULE_UNLOAD=y
-CONFIG_MODVERSIONS=y
```
ts-kris, line 66: Some configs keep modversions in, some remove it for some reason. These should be consistent.

ts-kris (commented):
I know this is urgent but there are a number of breaking things and inconsistencies in these that I noticed. I'd like to see them addressed in one way or another (either changes or clarifying comments) before I approve.

I recognize that a few of these options are likely removed because they are selected by another option, but they seem to be inconsistent in some places like support for something is enabled in one platform, but not another.

For a few of the comments, I tried to note if it was in multiple places or inconsistent across configs if I noticed it. Rather than repeating the same things over and over, I just noted in one comment that it affects multiple configs.

arch/arm/configs/tsimx6_defconfig
markfeathers: I'm going to add a requirement that CONFIG_EMBEDDED=n unless we find a strong reason to keep that.  I'll push new changes shortly for that.

arch/arm/configs/tsimx6_defconfig
markfeathers: Will fix.

arch/arm/configs/tsimx6_defconfig
markfeathers: I'll revisit this on our configs, I think we should be CONFIG_SLAB=y (not slob) unless we have a good reason.

arch/arm/configs/tsimx6_defconfig
markfeathers: I'll add this to our requirements to omit.

arch/arm/configs/tsa38x_defconfig
markfeathers: Fixed

arch/arm/configs/tsimx6_defconfig
markfeathers: Fixed

arch/arm/configs/tsimx6ul_defconfig
markfeathers: No, added to our requirements. Will fix.

arch/arm/configs/tsa38x_defconfig
markfeathers: Good catch, added these back to our requirements.

arch/arm/configs/tsimx6_defconfig
markfeathers: At this time none of the PC104 hosts covered by these configs support PC/104 ethernet cards.  They would all need a shim driver which does not yet exist.  This is also only possible on those that support using our memory windows on some platforms like the 6ul.

arch/arm/configs/tsimx6_defconfig
markfeathers: It looks like these are all not required by any drivers on this platform.

arch/arm/configs/tsimx6_minimal_defconfig
markfeathers: At this point I think all of our platforms need this so I'll add this to our requirements.

arch/arm/configs/tsimx6ul_defconfig
markfeathers: Added to requirements, surprised this one wasn't already there.

arch/arm/configs/tsimx6ul_defconfig
markfeathers: Added to our requirements.

arch/arm/configs/tsimx6ul_defconfig
markfeathers: Fixed

arch/arm/configs/tsimx6ul_defconfig
markfeathers: Afaik this uses CONFIG_RTC_DRV_SNVS, we should be able to skip MXC.

arch/arm/configs/tsimx6ul_defconfig
markfeathers: I'll add back a requirement for this.

arch/arm/configs/tsimx6_defconfig
markfeathers: Adding a requirement for this.

arch/arm/configs/tsimx6ul_defconfig
markfeathers: Adding requirements for this.  Minimal will be =n, standard will be =y.

luckshot (dismissed):

ts-kris (approved):
Everything looks good to me. Reviewed, not tested.

luckshot (approved):

PR #47, opened 2023-10-17
Update to 

markfeathers:
Sync with upstream


ts-kris:
I've not reviewed, but, did a quick glance over and the number of commits looks correct (plus your merge commit), and the hashes look like they were kept correctly as well.

```
$ git log --oneline HEAD...v5.10.170 | wc -l
3849

$ git log --oneline HEAD...v5.10.170 | head
a8d812240fdd Linux 5.10.198
660627c71bc1 xen/events: replace evtchn_rwlock with RCU
1e3d016a9506 mmc: renesas_sdhi: only reset SCC when its pointer is populated
a7d86a77c33b netfilter: nftables: exthdr: fix 4-byte stack OOB write
84f6b686df2d netfilter: nf_tables: fix kdoc warnings after gc rework
c17446c00805 parisc: Restore __ldcw_align for PA-RISC 2.0 processors
cfc333393ae6 RDMA/mlx5: Fix NULL string error
0d520cdb0cd0 RDMA/siw: Fix connection failure handling
5a4a6a47e074 RDMA/uverbs: Fix typo of sizeof argument
626868282c36 RDMA/cma: Fix truncation compilation warning in make_cma_ports
```

I think, moving forward, we need to somehow remove your current merge commit, a62da874c0cf10e43a14b1a946d077c4fad0fc16, as merging this PR would then create another merge commit. So I think rather than doing a merge and then a PR, we want to just open a PR with the new range of commits.

PR #63, opened 2024-01-10
ARM: dts: ts7100: LCD panel clock-frequency is 7 MHz (units: Hz)

lionel:


ts-kris (approved):
Tested.

markfeathers (approved):
Reviewed, not tested.

PR #66, opened 2024-01-23
ARM: dts: ts7180: fix merge typo

lionel:
Spotted this when diffing my local tree against 5.10.y after the merge. Figured it's better than to have a fix available to merge than to forget about it.


ts-kris (approved):
Yeah, I guess just "okay" is fine, but, double "okay" is even okay-er! ;)

markfeathers (approved):

PR #68, opened 2024-02-15
tsimx6ul: Use default kernel commandline in extend mode, use this to disable fbcon

ts-kris:
Note that while the diffs also show removal of `CONFIG_PPS` this is misleading as the option is still set to a module. It is selected by:
```
  │ Selected by [m]:
  │   - PTP_1588_CLOCK [=m] && NET [=y] && POSIX_TIMERS [=y]
```

Tested against WIP support for TS-7553-V2 in distro-seed.
Tested against TS-7100 Buildroot (LVGL demo).

In both cases the FB still functions, with no cursor blink or any other VT login text showing up on screen.


markfeathers (approved):
Reviewed, not tested.

randytsi369 (approved):

PR #75, opened 2024-04-09
pwm: pwm-ts: Fix inversed before enabling pwm

### LINKED ISSUE #74: pwm-ts: Allow setting polarity on boot without enabling
https://github.com/embeddedTS/linux-lts/blob/linux-5.10.y/drivers/pwm/pwm-ts.c#L140

We might need to implement the hooks:
disable
set_polarity

Or "apply" might be called when we are disabled, which appears to be the case since we handle disabling the state in apply. We should determine if we can implement a default idle state before a cycle is issued.

As a workaround for now, it looks like enabling a period/duty cycle with a polarity, and then disabling it would set an idle state.

markfeathers:
This allows inversed to be set without first turning on the PWM. This also removes the state in the ts object that was tracking state, and now the calc function sets the registers. Removed the spinlock which is not used or needed.

### DIFF
--- a/drivers/pwm/pwm-ts.c
+++ b/drivers/pwm/pwm-ts.c
@@ -16,7 +16,6 @@
 #include <linux/pm_runtime.h>
 #include <linux/pwm.h>
 #include <linux/slab.h>
-#include <linux/spinlock.h>
 
 /* Enabled bit is the only one that applies immediately.  All other registers
  * take effect when apply is set
@@ -37,11 +36,6 @@
 struct ts_pwm {
 	struct pwm_chip chip;
 	void __iomem *base;
-	spinlock_t lock;
-	u16 duty;
-	u16 period;
-	u8 shift;
-	struct pwm_state state;
 	struct clk *clk;
 };
 
@@ -56,7 +50,9 @@ static int ts_pwm_calc(struct ts_pwm *ts,
 {
 	unsigned long clk_rate = clk_get_rate(ts->clk);
 	unsigned long long cycle;
-	unsigned int shift, cnt, duty_cnt;
+	unsigned int  cnt, duty_cnt;
+	u16 duty_reg;
+	u8 shift;
 
 	/* Calc shift & period reg */
 	for (shift = 0; shift < SHIFT_MAX; shift++) {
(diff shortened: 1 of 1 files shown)


randytsi369 (dismissed):
Reviewed, not tested.

luckshot (dismissed):

ts-kris:
I havn't forgot about this, just had higher priority things on my plate. If this is super urgent I can cut over and test

drivers/pwm/pwm-ts.c
```diff
@@ -90,49 +83,36 @@ static int ts_pwm_calc(struct ts_pwm *ts,
...
-
-		ts->shift = shift;
-		ts->period = cnt;
-		ts->duty = cnt - duty_cnt;
+		duty_reg = cnt - duty_cnt;
 	}
 
+	writew(cnt, ts->base + REG_PERIOD);
+	writew(duty_reg, ts->base + REG_DUTY);
+	writew(shift, ts->base + REG_SHIFT);
```
ts-kris, line 91: I'm curious what this looks like on the PWM line when there is a change? In looking at other PWM drivers, it looks like in general they rely on hardware to only apply settings at a specific time. I'm not sure what our core does in this case. If it is really glitchy, we might consider disabling the PWM, applying settings, and then enabling the PWM?

drivers/pwm/pwm-ts.c
```diff
@@ -90,49 +83,36 @@ static int ts_pwm_calc(struct ts_pwm *ts,
...
-		return 0;
-	}
+	u16 ctrl = 0;
 
 	if (state->polarity == PWM_POLARITY_NORMAL)
-		value &= ~(INVERSED);
+		ctrl &= ~INVERSED;
 	else
-		value |= INVERSED;
+		ctrl |= INVERSED;
```
ts-kris: This can be just
```
if (state->polarity != PWM_POLARITY_NORMAL)
		ctrl |= INVERSED;
```

No need for the if/else

markfeathers: Good point, will fix.

drivers/pwm/pwm-ts.c
markfeathers: The PWM framework enforces that the period does not change when the PWM channel is enabled. I based this on pwm-mxs which does seem to write its own registers similarly. The period changing is all that can influence our period/shift registers having different values, so the only piece that can change is duty. The FPGA core is safe to change duty at any time. I could update this in code to check if its enabled, but I'd have to store the cnt/cycle values from when period actually is set. Its certainly doable to store & have different paths for if its enabled or not to save the extra calculation / write when the user is only changing duty, but I'm tempted to leave that until we have a need to change duty faster.

ts-kris: No, no, that should be okay.
> The PWM framework enforces that the period does not change when the PWM channel is enabled.

I was unaware of that, and as shift is only affected by period, then only duty would ever change and all of this should be safe.

ts-kris (approved):

randytsi369 (approved):
Reviewed, not tested.

PR #78, opened 2024-04-23
ARM: dts: imx28: fix enet flapping issue

ts-kris:
I want to highlight a couple things about this moving forward.

Any time we use a CPU clock to drive a PHY, this clock _needs_ to be added to the `ethphy` handle. It looks like we do this in some places already. Any new devicetrees need to have this.


arch/arm/boot/dts/imx28-ts7670.dts
randytsi369, line 1: Typo in comment: "magic number"

arch/arm/boot/dts/imx28-ts7400v2.dts
randytsi369, line 1: Typo in comment: "magic number"  Line 213

randytsi369 (approved):
Reviewed, no issues found. Not tested.

markfeathers (approved):
Reviewed, not tested.

PR #79, opened 2024-04-24
ARM: dts: imx6qdl-ts7990: Add i2c pinctrl and sda/scl gpios for recovery

markfeathers:
Set default speed, and added i2c recovery gpio.


randytsi369 (approved):
Reviewed, not tested.

ts-kris (approved):
Reviewed, not tested

PR #92, opened 2024-06-21
fix yellow and blue LED gpio

dataselfservice:
fixing DTS incorrect GPIOs for yellow and blue LEDs

### DIFF
--- a/arch/arm/boot/dts/imx28-ts7680.dts
+++ b/arch/arm/boot/dts/imx28-ts7680.dts
@@ -49,14 +49,14 @@
 		led-2 {
 			color = <LED_COLOR_ID_YELLOW>;
 			function = LED_FUNCTION_INDICATOR;
-			gpios = <&gpio1 26 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio1 58 GPIO_ACTIVE_LOW>;
 			default-state = "off";
 		};
 
 		led-3 {
 			color = <LED_COLOR_ID_BLUE>;
 			function = LED_FUNCTION_INDICATOR;
-			gpios = <&gpio1 24 GPIO_ACTIVE_HIGH>;
+			gpios = <&gpio1 56 GPIO_ACTIVE_HIGH>;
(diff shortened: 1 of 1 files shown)


ts-kris:
I do not believe this PR is correct. I'll have to go back and test it but that won't be till next early week as I'm in the middle of other tasks.

My notes show I've tested the LEDs and they are all good. Additionally, the &gpio1 bank only has 32 GPIOs, so 58 and 56 are not valid pins in that bank.

Is there a specific issue you are seeing? If you prefer, you are welcome to reach out to on our support channels for a private conversation on the matter.

dataselfservice:
Hi,

LEDs were not working by accessing sysfs gpio 58, 56. After patch they
work. Might be still something to tune if bank in 32 and 56 58 overflow.

ts-kris:
Ah, I understand the confusion.

The short of it is, the 5.10 port of the TS-7680 connects more parts of the hardware to common kernel interfaces that were either non-existent or incomplete in earlier kernel versions. Part of this port will see a documentation update for the TS-7680, however, this may be a little bit out still as we are working out details of separate hardware and software manuals as platforms in our LTS group have spanned many kernel releases and old paradigms are no longer used.

The LED subsystem is now a separate entity and can be accessed via `/sys/class/leds/` Each LED has a folder within that, and can be controlled manually via the `brightness` file, or various triggers can be set via the `trigger` file (e.g. CPU activity, disk activity, etc.; `cat` the `trigger` file to see the currently set trigger and all available triggers). More detail can be found here <https://docs.kernel.org/leds/leds-class.html>

Pre-patch, you are unable to open the LEDs by direct GPIO number because they are consumed by the LED system. The patch worked because you assigned the LEDs to a non-existent GPIO, which allowed you to open the GPIO directly.

We recommend moving your application and use to use the LED subsystem, or, if you prefer to directly use GPIOs, then you need to separately patch your local tree.

dataselfservice:
Crystall clear. Thanks

PR #96, opened 2024-07-15
Update to v6.6.40

DaftB0t:
Sync with upstream v6.6.40


ts-kris:
Do not merge yet, still testing automations.

PR #102, opened 2024-07-18
serial: 8250: 8250_ts: Enable RS-485 through RTS

markfeathers:
This adds support for RS-485 on our PC/104 uart devices by having the kernel automatically control RTS around transactions.


ts-kris (approved):
Not tested, looks good to me.

randytsi369 (approved):
Looks good to me, not tested.

PR #181, opened 2025-02-19
6.6.y: i.MX28 support

### LINKED ISSUE #109: TS-7680: linux-6.6.y: Unable to reboot the system
The TS-7680 imported in to 6.6.y mostly cleanly, however reboot functionality has broken.

Some history/notes:
In 4.9, the `ts_wdt` driver would register itself as the highest priority reset handler when it loaded. In 5.10 however, some changes to the machine files for mxs CPUs caused the kernel to use the mxs registered restart handler and bypasses the notifier/restart chain. This overrode the `arm_pm_restart` variable that the mxs machine file set. In 6.6, neither of these work for some reason. The Arm restart process no longer uses `arm_pm_restart` and will always use the restart notifier chain. However, despite `ts_wdt` registering itself as a restart handler, at the highest priority, the WDT restart handler never gets called.

Eventually, the platform will reboot with the WDT timing out as it is not closed at poweroff.

### LINKED ISSUE #126: TS-7670, TS-7400-V2: 6.6: Ethernet flapping issue returns
Despite finding the root cause and implementing a proper fix in 5.10 (0e138dfe10981e1064f76bd724883dde067a141e), something else changed with the PHY setup in 6.6. Unsure yet if it is the clock handling, or maybe reset control, more research is needed.

Note that, it was also found that in 6.6, the SMSC PHY driver was built as a module and not being loaded. Adding the driver as a builtin allowed for proper PHY detection, however, the issue still remained.

ts-kris:
Adds support for TS-7670, TS-7400-V2, and TS-7680

Previously, the TS-7670 and TS-7400-V2 saw the return of the ethernet flapping issue, #126. This has been resolved.

Additionally the TS-7680 had an issue with rebooting, #109. This has also been resolved.

On all three platforms this series also relocates the I2C0 bus from CPU peripheral to GPIO. This is due to the MXS I2C driver not offering an atomic set of xmit functions which resulted in a handful of kernel warnings on shutdown/reboot since the microcontroller WDT is used as part of the reboot function (and stopped for a halt). While not an issue since everything worked, removing the warning looks nicer.

This also applies some other dts changes to clean up improperly set up `memory` nodes.


ts-kris:
The reported failing checks appear to be related to TS-7680 specifically. I renamed some of the node names to more correctly match devicetree paradigms and that is causing failures of the tests. I'll address this shortly.

markfeathers (approved):
The changes look good to me as long as the tests are updated & pass before merged.

ts-kris:
Tests passing, PR opened for CI system to get those changes merged in.

luckshot (approved):

PR #188, opened 2025-03-21
TS-7100 touchscreen updates

markfeathers:
This fixes the touchscreen on the 6.6 kernel


randytsi369 (dismissed):
Reviewed, not tested.

ts-kris (approved):
Confirmed with Mark that with the FPGA change to accommodate this for 6.6, the codepath for 5.10 still works and will not need to be modified.

luckshot (approved):

PR #199, opened 2025-05-02
Update to v6.6.89

DaftB0t:
Sync with upstream v6.6.89


ts-kris:
The issue here is a conflict against 780d24d054521c795cc2c0b3b9a0bb869ff40d5f (Revert "usb: chipidea: ci_hdrc_imx: decrement device's refcount in .remove() and in the error path of .probe()")

780d24d054521c795cc2c0b3b9a0bb869ff40d5f reverts dcd4de31bd01a7189c24e3cafe40649c9c42b9af since the addition of dcd4de31bd01a7189c24e3cafe40649c9c42b9af caused i.MX28 platforms to catch a NULL pointer dereference on driver teardown since it appeared to be decrementing a refcount that doesn't seem to be incremented anywhere despite dcd4de31bd01a7189c24e3cafe40649c9c42b9af saying it is.

This warrants further investigation and maybe a proper fix due to this conflict. I may just rip out 780d24d054521c795cc2c0b3b9a0bb869ff40d5f, get this merged, and then open an Issue to try and better resolve the actual problem.

PR #203, opened 2025-05-16
feat: added led labels to TS-7970 FDT

dataselfservice:
This if securing compliance with TS-7970 manual
<https://docs.embeddedts.com/TS-7970#LEDs> LEDs naming: `/sys/class/leds/{blue,green,yellow-red}-led/`, instead of current kernel auto-naming scheme which might change due to kernel changes.

### DIFF
--- a/arch/arm/boot/dts/imx6qdl-ts7970.dtsi
+++ b/arch/arm/boot/dts/imx6qdl-ts7970.dtsi
@@ -23,6 +23,7 @@
 		compatible = "gpio-leds";
 
 		led-0 {
+			label = "green-led";
 			color = <LED_COLOR_ID_GREEN>;
 			function = LED_FUNCTION_POWER;
 			gpios = <&gpio3 27 GPIO_ACTIVE_LOW>;
@@ -30,6 +31,7 @@
 		};
 
 		led-1 {
+			label = "red-led";
 			color = <LED_COLOR_ID_RED>;
 			function = LED_FUNCTION_STATUS;
 			gpios = <&gpio1 2 GPIO_ACTIVE_LOW>;
@@ -37,6 +39,7 @@
 		};
 
 		led-2 {
+			label = "yellow-led";
 			color = <LED_COLOR_ID_YELLOW>;
 			function = LED_FUNCTION_INDICATOR;
 			gpios = <&gpio1 9 GPIO_ACTIVE_LOW>;
@@ -44,6 +47,7 @@
 		};
 
 		led-3 {
+			label = "blue-led";
 			color = <LED_COLOR_ID_BLUE>;
 			function = LED_FUNCTION_INDICATOR;
 			gpios = <&gpio4 25 GPIO_ACTIVE_HIGH>;


markfeathers:
Hi!

There have been some changes since the initial TS-7970 software releases. We've since been trying to upstream more of our board support, and some of the feedback for best practices we've received is instead using the function/color rather than naming each led by their color. Unfortunately [labels are deprecated](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/leds/common.yaml?h=v6.15-rc6#n54).

We're working on updating manuals to solve this issue in general. Having one single manual that handles all of the possible Linux releases ends making it very complicated to read. Different kernels have changed things like sysfs gpio to gpiod, and different linux images require different behaviors for things like x11/wayland, or even bluetooth startup. We're working on testing out a split between a hardware manual and having software manuals that go with each major Linux release. This is being tested out on another platform first and is not yet done, but we're hoping to eventually bring this to all of our platforms supported in linux-lts.

In the meantime, let us know through our [ticketing system](https://support.embeddedts.com/support/home) / support@embeddedTS.com if any other new-kernel behaviors come up and we can help clarify. The most common one has been GPIO changing its numbering, but "gpioinfo" on the board includes the net names as a lookup which should help provide a mapping.

dataselfservice:
The resulting LED naming is inconsistent with your documentation. Also it appears that kernel 5.10.y is handling naming a bit in a wierd way: `/sys/class/leds/blue:` (note `:`)?!?! Turning to 5.15.y (nxp-imx/linux-imx) is uses node name `led-0`. It is a bit of mess. If labels are not an option, what do you suggest to achieve predictable naming?

ts-kris:
> The resulting LED naming is inconsistent with your documentation. 

As Mark noted, having consistent documentation across many major releases of kernel is difficult and something we are working on addressing by having documentation that is centered on specific major releases. Until then, however, we are happy to answer any questions arising from confusion as kernel paradigms and processes change in major releases.

> Also it appears that kernel 5.10.y is handling naming a bit in a wierd way: `/sys/class/leds/blue:` (note `:`)?!?!

If this is the case, this is indeed an issue. I will look in to this first thing Monday. Can you let me know what specific platform and image date you are using? If you do not wish to share this publicly, you are encouraged to open a support ticked through our [ticketing system](https://support.embeddedts.com/support/home) or send an email to support@embeddedTS.com. You can address it to me, Kris, and I will make sure it gets looked at ASAP.

> Turning to 5.15.y (nxp-imx/linux-imx) is uses node name led-0. It is a bit of mess.

Unfortunately that is not a kernel we are involved in. With our linux-lts project we are aiming to get as close as possible to vanilla Linux requirements for specific releases. 

> If labels are not an option, what do you suggest to achieve predictable naming?

You are always welcome to fork the kernel and add a label as a patch yourself. Additionally, if you are using any of our build tools ([distro-seed](https://github.com/embeddedTS/distro-seed/), [Buildroot-ts](https://github.com/embeddedTS/buildroot-ts), or [Yocto BSP](https://github.com/embeddedTS/ts-oe-bsp) / [Yocto meta-ts](https://github.com/embeddedTS/meta-ts)) you can use these to apply patches to the kernel tree. This would allow you to maintain a specific devicetree or patch to devicetree file in your build system to apply a unique patch.

Additionally, aside from changes due to bugfixes (e.g. if we need to change the devicetree to adjust the `blue` LED filename) major kernel releases will remain unchanged in ways that would break userspace. For example, the 4.9 kernel for the TS-7970 used labels, but the move to 5.10.y changed to the `color:function` format. The 4.9 will not have this change backported.

dataselfservice:
@ts-kris Noted. We can still override the `led-controller` node in FDT. Thanks for the further clarification.

PR #216, opened 2025-07-01
ARM: dts: imx6ul: ts7553v2: Add Rev E support

markfeathers:
Adds 2 new nimbelink control pins
Changes from mpu9250a to st,lis2mdl/st,ism330dhcx.

### DIFF
--- a/arch/arm/boot/dts/nxp/imx/imx6ul-ts7553v2.dts
+++ b/arch/arm/boot/dts/nxp/imx/imx6ul-ts7553v2.dts
@@ -126,6 +126,17 @@
 		sda-gpios = <&gpio5 5 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 		scl-gpios = <&gpio5 4 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 
+		lis2mdl: magnetometer@1e {
+			compatible = "st,lis2mdl";
+			reg = <0x1e>;
+			pinctrl-names = "default";
+			pinctrl-0 = <&pinctrl_magn_gpio>;
+			st,drdy-int-pin = <1>;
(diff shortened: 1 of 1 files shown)


randytsi369 (dismissed):
reviewed, not tested.

arch/arm/boot/dts/nxp/imx/imx6ul-ts7553v2.dts
```diff
@@ -126,6 +126,17 @@
 		sda-gpios = <&gpio5 5 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 		scl-gpios = <&gpio5 4 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 
+		magnet: magnetometer@1e {
```
ts-kris:
```suggestion
		lis2mdl: magnetometer@1e {
```

arch/arm/boot/dts/nxp/imx/imx6ul-ts7553v2.dts
```diff
@@ -436,6 +457,12 @@
 		>;
 	};
 
+	pinctrl_gyro_gpio: gyrogrp {
+		fsl,pins = <
+			MX6UL_PAD_LCD_ENABLE__GPIO3_IO01	0x13020
+		>;
+	};
+
```
ts-kris, line 465: Does moving this pin, IMU IRQ on previous PCB revs, still function correctly on older PCB revs? Since this pin could now be used by the new or old device, I think this would be better to leave with the I2C iomux group since it could be used by either downstream device.

ts-kris (changes requested):

ts-kris (approved):

randytsi369 (approved):
Reviewed, not tested.

PR #217, opened 2025-07-01
ARM: dts: imx6ul: ts7553v2: Add Rev E support

markfeathers:
Adds 2 new nimbelink control pins
Changes from mpu9250a to st,lis2mdl/st,ism330dhcx.

### DIFF
--- a/arch/arm/boot/dts/imx6ul-ts7553v2.dts
+++ b/arch/arm/boot/dts/imx6ul-ts7553v2.dts
@@ -126,6 +126,17 @@
 		sda-gpios = <&gpio5 5 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 		scl-gpios = <&gpio5 4 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 
+		lis2mdl: magnetometer@1e {
+			compatible = "st,lis2mdl";
+			reg = <0x1e>;
+			pinctrl-names = "default";
+			pinctrl-0 = <&pinctrl_magn_gpio>;
+			st,drdy-int-pin = <1>;
+			interrupt-parent = <&gpio1>;
+			interrupts = <13 IRQ_TYPE_LEVEL_HIGH>;
+		};
+
+		/* Only present on REV D and below */
 		mpu9250a: imu@68 {
 			compatible = "invensense,mpu9250";
 			reg = <0x68>;
@@ -139,6 +150,13 @@
 					reg = <0xc>;
 				};
 			};
+		};
+
+		ism330: gyro@6a {
+			compatible = "st,ism330dhcx";
+			reg = <0x6a>;
+			interrupt-parent = <&gpio3>;
+			interrupts = <1 IRQ_TYPE_LEVEL_HIGH>;
 		};
 	};
 };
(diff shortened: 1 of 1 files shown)


randytsi369 (dismissed):

arch/arm/boot/dts/imx6ul-ts7553v2.dts
```diff
@@ -126,9 +126,22 @@
 		sda-gpios = <&gpio5 5 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 		scl-gpios = <&gpio5 4 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN)>;
 
+		magnet: magnetometer@1e {
```
ts-kris:
```suggestion
		lis2mdl: magnetometer@1e {
```

arch/arm/boot/dts/imx6ul-ts7553v2.dts
```diff
@@ -454,7 +476,18 @@
...
 			MX6UL_PAD_SNVS_TAMPER5__GPIO5_IO05	0x4001a8b0
-			MX6UL_PAD_LCD_ENABLE__GPIO3_IO01	0x13020 /* IMU IRQ */
+		>;
+	};
+
+	pinctrl_gyro_gpio: gyrogrp {
+		fsl,pins = <
+			MX6UL_PAD_LCD_ENABLE__GPIO3_IO01	0x13020
+		>;
+	};
```
ts-kris: Does moving this pin, IMU IRQ on previous PCB revs, still function correctly on older PCB revs? Since this pin could now be used by the new or old device, I think this would be better to leave with the I2C iomux group since it could be used by either downstream device.

ts-kris (changes requested):

arch/arm/boot/dts/imx6ul-ts7553v2.dts
markfeathers: This should work, but I'm unable to test the old part at this time because its not on the test rack. Linux resolves pinctrl states when the respective driver loads.  Since Linux doesn't load the driver until the correct device acks on i2c, it doesn't ever try to load both pinctrls at once. I can say its working correctly on the new rev and there are no iomux errors from it trying to have multiple drivers claim the same pin, but since I can't test the old implementation I'll roll this back.

ts-kris: >  Since Linux doesn't load the driver until the correct device acks on i2c, it doesn't ever try to load both pinctrls at once.

I wasn't completely certain how that worked, and less certain about ordering and when drivers claim pins vs how IOMUX can be affected. Based on what you said, if on an old rev the old driver attempted to probe first, and succeeded, then the newer driver loaded, the first driver already claimed the interrupt pin so the newer driver fails to set IOMUX.

I'm also not certain offhand it if would create conflicts if each driver had a reference to that IOMUX node handle.

I think it is just easier to keep the shared INT as part of the I2C node rather than as part of any device node.

arch/arm/boot/dts/imx6ul-ts7553v2.dts
ts-kris: Looks like this got clobbered after your last push.

ts-kris (approved):

randytsi369 (approved):
Reviewed, not tested.

PR #225, opened 2025-07-17
ARM: embeddedts: rename tssupervisor subsystem to tswizard

luckshot:
This series completes the transition from the legacy "tssupervisor" naming
convention to the updated "tswizard" naming used internally at embeddedTS.

Changes include:
  - Renaming all embeddedTS-specific driver sources and headers:
      drivers/mfd/tssupervisor_core.c  → drivers/mfd/tswizard_core.c
      drivers/iio/adc/tssupervisor_adc.c → drivers/iio/adc/tswizard_adc.c
      drivers/iio/temperature/tssupervisor_temp.c → drivers/iio/temperature/tswizard_temp.c
      drivers/rtc/rtc-tssupervisor.c → drivers/rtc/rtc-tswizard.c
      drivers/power/reset/tssupervisor-reset.c → drivers/power/reset/tswizard-reset.c

  - Updating corresponding Kconfig and Makefile entries

  - Renaming device tree bindings where applicable

  - Replacing all `CONFIG_*TSSUPERVISOR*` entries with equivalent
    `CONFIG_*TSWIZARD*` symbols

  - Updating `tsimx6ul_defconfig` and `tsimx6ul_minimal_defconfig` to
    enable the full tswizard driver stack:
      - CONFIG_MFD_TS_WIZARD
      - CONFIG_TS_WIZARD_ADC
      - CONFIG_TS_WIZARD_TEMP
      - CONFIG_POWER_RESET_TS_WIZARD
      - CONFIG_RTC_DRV_TSWIZARD

All changes are restricted to embeddedTS-specific drivers and board configs.
Builds for affected targets (`ts7250v3`, `ts7180`, `tsimx6ul`, etc.) have
been validated with `make olddefconfig` and `make savedefconfig` to ensure
correct integration.

This rename aligns kernel naming with current product documentation and
internal conventions, reducing confusion and improving consistency across
the codebase.

Signed-off-by: Michael Peters <mpeters@embeddedts.com>


arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
```diff
@@ -155,18 +155,18 @@
 		status = "disabled";
 	};
 
-	supervisor: supervisor@10 {
-		compatible = "technologic,supervisor";
+	wizard: wizard@10 {
+		compatible = "technologic,tswizard";
```
ts-kris: I see other mentions of `wizard-adc` `wizard-temp` etc., this is the first use of `tswizard`. Since we didn't use `tssupervisor`, i.e. the change is from `supervisor`, the change should be to `wizard` for consistency. Either that, or other instances of `wizard` like `-adc` should probably move to `tswizard-adc` and `tswizard-temp`. Though I prefer dropping the `ts` on there since the full string already includes the manufacturer, e.g. `technologic,wizard`

drivers/iio/adc/tswizard_adc.c
```diff
@@ -149,21 +149,21 @@ static int ts_supervisor_adc_probe(struct platform_device *pdev)
 	return devm_iio_device_register(&pdev->dev, indio_dev);
 }
 
-static const struct of_device_id tssupervisor_of_match[] = {
-	{ .compatible = "technologic,tssupervisor-adc", },
+static const struct of_device_id tswizard_of_match[] = {
+	{ .compatible = "technologic,tswizard-adc", },
```
ts-kris, line 153: This seems to be a potential issue. I noted in my previous comment that `technologic,wizard-adc` was used, but this looks like its matching `technologic,tswizard-adc`. This should be rectified. But is also strange because this change goes from `technologic,tssupervisor-adc` but the change above in the devicetree changed from `technologic,supervisor-adc`.

drivers/mfd/tswizard-core.c
```diff
@@ -7,28 +7,28 @@
...
 	{
-		.name = "tssupervisor-temp",
-		.of_compatible = "technologic,supervisor-temp",
+		.name = "tswizard-temp",
+		.of_compatible = "technologic,tswizard-temp",
 		.id = -1,
 	},
 	{
-		.name = "tssupervisor-adc",
-		.of_compatible = "technologic,supervisor-adc",
```
ts-kris, line 26: I see this change goes from `technologic,supervisor-adc`

It looks like drivers/iio/adc/tswizard_adc.c has a conflicting compatible string, so that went from `technologic,tssupervisor-adc` to `technologic,tswizard-adc` but this is going from `technologic,supervisor-adc` to `technologic,tswizard-adc`

Based on this, it looks like we had very fragile naming, two drivers, one using `supervisor-adc` and another one using `tssupervisor-adc`, same with `-temp`. This should be looked at closer so we don't have conflicting or fragile compatible names.

drivers/mfd/tswizard-core.c
```diff
@@ -291,44 +291,45 @@ static int ts_supervisor_i2c_probe(struct i2c_client *client)
...
 	{ /* sentinel */ }
 };
-MODULE_DEVICE_TABLE(i2c, ts_supervisor_i2c_id);
+MODULE_DEVICE_TABLE(i2c, ts_wizard_i2c_id);
 
-static const struct of_device_id ts_supervisor_i2c_of_match[] = {
-	{ .compatible = "technologic,supervisor", },
+static const struct of_device_id ts_wizard_i2c_of_match[] = {
+	{ .compatible = "technologic,tsupervisor", },
+	{ .compatible = "technologic,tswizard" },
```
ts-kris: Leaving old compatible in is a great idea, I didn't notice any other drivers doing that. Is that something we should be doing across all drivers? And also, this breaks the old compatible string anyway, so, this may not be as intended?

ts-kris (changes requested):

drivers/mfd/tswizard-core.c
markfeathers: There ended up not being a good way to do that as we looked at it more. The MFD driver has to advertise what the sub devices are in C code, and list the matching compatible strings.  There is only 1 compatible string you can list, so we can't support both without maintaining two entirely separate lists of mfd devices.  IMO we should just suggest the change at that point.

I'd also suggest we just remove tssupervisor from here even if we aren't going to support that.

luckshot:
🚧 Force-pushed update — addressed prior review concerns.

This push completes the renaming effort from tswizard to wizard for internal code identifiers, while preserving external-facing names used for build integration and compatibility.

✅ Summary of changes:

Renamed internal identifiers such as ts_wizard, tswizard_, and TS_WIZARD_MAX_ADC to wizard, wizard_, and WIZARD_MAX_ADC respectively.

Updated struct names, variables, function names, and comments to drop the ts prefix internally.

Preserved all external-facing names, including:

File names (e.g. tswizard_adc.c)

Kconfig symbols (e.g. CONFIG_TS_WIZARD_*)

Makefile references

Updated DTS-compatible strings from "technologic,tswizard" to "technologic,wizard" as requested.

Ensured all #include paths referencing ts_wizard.h or related headers remain unchanged.

✅ Build tested: make completes cleanly with all modules building successfully.

Let me know if anything else needs adjustment—this should resolve the blocking feedback.

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
luckshot: The original strategy was to exchange like-for-like, but it turned out the original code uses nearly all permutations of ts and supervisor with and without caps, hyphens, and underscores.  So the strategy was tswizard would be used to define our namespace, and wizard would be used where the namespace is already defined.

drivers/mfd/tswizard-core.c
luckshot: This *should* now be resolved by using tswizard-*  for all compatible strings. New commit posted.

drivers/mfd/tswizard-core.c
luckshot: I think it's a great idea but on review it seemed like it wasn't universally possible.

luckshot: OK removing the extra .compatible, it'll be in my next force-push.

luckshot: I seem to be unable to resolve this conversation... do not know why.

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
```diff
@@ -111,20 +111,13 @@
...
 		io-channel-names = "3.3V", "VDD_ARM_CAP", "VDD_SOC_CAP", "5V_A", "8V_48V";
 		status = "disabled";
 	};
-
-	supervisor-temp {
-		compatible = "iio-hwmon";
-		io-channels = <&supervisor_temp>;
-		io-channel-names = "Supervisor Temp";
-		status = "okay";
-	};
```
ts-kris: This disappeared without a replacement, should this be re-created under iiomon?

drivers/iio/temperature/tswizard_temp.c
ts-kris, line 1: Not sure if its an artifact on how the change was made, but it looks like this should be moved from `tssupervisor_temp.c` but github thinks that was a deleted file and this is a new file.

ts-kris (dismissed):

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
luckshot: This node was causing the kernel to create a platform_device before the MFD driver started. When our MFD driver then tried to create its own wizard-temp child device, it caused a "duplicate filename" error, which halted the parent driver's probe.

markfeathers (dismissed):

drivers/mfd/Kconfig
lionel, line 1: We accidentally modified line 2279 in the `RAVE_SP_CORE` config. We should remove that change.

luckshot: Fixed. Changed s to S in supervisor on RAVE_SP_CORE description.

markfeathers (approved):

randytsi369 (approved):
Reviewed, not tested.

PR #235, opened 2025-09-09
i2c-gpio for ts-silo-104 support on ts-7250-v3

luckshot:
Adds device tree overlay for i2c-gpio support on the PC/104 header on the TS-7250-V3.


arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
```diff
@@ -34,7 +34,6 @@
 		gpio9 = &fpga_bank4;
 		gpio10 = &fpga_bank5;
 		gpio11 = &pc104gpio;
-		i2c5 = &i2c5_gpio;
```
ts-kris: Changes to this file are basically reverting the previous commit, 18e58a0998d76abf8e26dce446128e543851366c. If the 18e58a0998d76abf8e26dce446128e543851366c commit is not necessary, it should be squashed so that the lines are not added then removed in two back to back commits

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
```diff
@@ -46,15 +45,6 @@
...
 
-	// pc104 pins B15 and B17.
-	i2c5_gpio: i2c-104 {
-		compatible = "i2c-gpio";
-		scl-gpios = <&fpga_bank1 13 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
-		sda-gpios = <&fpga_bank1 8 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
-		#address-cells = <1>;
-		#size-cells = <0>;
-	};
-
```
ts-kris: Changes to this file are basically reverting the previous commit, 18e58a0998d76abf8e26dce446128e543851366c. If the 18e58a0998d76abf8e26dce446128e543851366c commit is not necessary, it should be squashed so that the lines are not added then removed in two back to back commits

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi
```diff
@@ -91,7 +81,7 @@
 
 	an_3p3v: an_3p3v {
 		compatible = "voltage-divider";
-		io-channels = <&supervisor_adc 0>;
+		io-channels = <&wizard_adc 0>;
```
ts-kris: Changes from `supervisor` to `wizard` should have been fully covered in #225. If these were missed (it happens), they should be redone as a different PR, no need to pollute this PR with other changes.

This applies to multiple changes in this file (`arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtsi`) as well as `arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dts`

arch/arm/boot/dts/nxp/imx/Makefile
```diff
@@ -411,10 +420,5 @@ dtb-$(CONFIG_SOC_IMX7D) += \
 dtb-$(CONFIG_SOC_IMX7ULP) += \
 	imx7ulp-com.dtb \
 	imx7ulp-evk.dtb
-dtbo-$(CONFIG_SOC_IMX6UL) += imx6ul-ts7250v3-silo104-i2c-gpio.dtbo
 dtb-$(CONFIG_SOC_IMXRT) += \
 	imxrt1050-evk.dtb
-
-# Allow overlays to be sourced from .dtso files too
-$(obj)/%.dtbo: $(src)/%.dtso FORCE
-	$(call if_changed,dtc_overlay)
```
ts-kris: It looks like this is reverting changes added in 8d49c841657eeb58f9640495e6a5529ae184eb91. If the 8d49c841657eeb58f9640495e6a5529ae184eb91 commit was not necessary, these should be squashed so only the singular needed change is made as opposed to multiple commits back to back adding and then removing code.

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtso
```diff
@@ -0,0 +1,36 @@
...
+
+                /* Open-drain for both lines (legacy property name for broad kernel support) */
+                i2c-gpio,open-drain;
+
+                /* Optional timing tweak: */
+                // i2c-gpio,delay-us = <5>;
+
+                #address-cells = <1>;
+                #size-cells = <0>;
+                status = "okay";
```
ts-kris: `okay` is the default status. While I'm not familiar enough with devicetree overlays, that should still apply here and the status line should not be needed.

arch/arm/boot/dts/nxp/imx/imx6ul-ts7250v3.dtso
```diff
@@ -0,0 +1,36 @@
...
+            };
```
ts-kris: Technically, an unused bus without a device is considered invalid/dead code and shouldn't be present. It can be marked as `status = "disabled";` if it is a framework that another devicetree can then add a device sub-node and enable the bus.

That also depends on the device having a driver and not relying on `i2cdev`, so, if we are relying on a userspace application to talk to an i2cdev device then this is valid. But if we are connecting this to a kernel driver, and this PR is meant to fully introduce support for the device, then the device should be present as a subnode of the I2C bus here.

ts-kris (changes requested):

ts-kris:
Github is also saying this branch is out of date, part of the reason there are lingering `supervisor` to `wizard` conversions may be because you need to rebase the `gpio_i2c_6.6.y` branch against the latest `linux-6.6.y` branch that includes #225.

PR #240, opened 2025-09-19
drivers: irqchip: ts71xxweim: Implement ack

markfeathers:
Older FPGAs would sort of "ack" on register read, but this doesn't necessarily agree with how Linux expects IRQs. This results in some IRQs being dropped, and some IRQ X and nobody cared messages. Newer FPGA releases include a new ack enable bit, which changes the behavior and the IRQ will then use a normal ACK.


luckshot (approved):

markfeathers:
This is handled in #289

PR #288, opened 2026-03-27
spi: oc: Driver rework following kernel norms

markfeathers:
Fixed dts bindings to use standard "num-cs"
Fixed dts binding to remove extra index
Fixed deprecated SPI queue
Removed dead PM code
Reworked around new transfer_one hooks
Fixed support for cs-gpios
Fixed RX only to TX 0s
Removed raw bit operators and used helpers where possible


ts-kris (changes requested):
This doesn't apply the changes to TS-7100 dtsi

ts-kris (approved):

PR #275, opened 2026-02-13
ts71xxweim: extend IRQ functionality for interrupt controller and gpio controller

anakin-childerhose-sfl:
# irq-ts71xxweim changes

Changes made:
- add ack and mask_ack functions
- add spinlocks for read-modify-write transactions
- add initialisation of priv->mask to value of mask register
- simplified irq handler
- minor fixes

These commits have been tested on a ts7250 v3 with FPGA firmware rev70 by jumping the RX and TX pins together for the three RS-232 UARTS and running them all at 115200 baud for a period of time:
- ttyS8 on the DB9 connector
- ttyS9 on COM2 header
- ttyS10 on COM3 header

The output of linux-serial-test for each tty:
```
/dev/ttyS10: count for this session: rx=3919758, tx=3919814, rx err=0
/dev/ttyS10: TIOCGICOUNT: ret=0, rx=3997939, tx=3997939, frame = 0, overrun = 0, parity = 0, brk = 0, buf_overrun = 0
/dev/ttyS8: count for this session: rx=3923653, tx=3923653, rx err=0
/dev/ttyS8: TIOCGICOUNT: ret=0, rx=4009904, tx=4009904, frame = 0, overrun = 0, parity = 0, brk = 0, buf_overrun = 0
/dev/ttyS9: count for this session: rx=3929020, tx=3932100, rx err=0
/dev/ttyS9: TIOCGICOUNT: ret=0, rx=4018415, tx=4018415, frame = 0, overrun = 0, parity = 0, brk = 0, buf_overrun = 0
```

# gpio-ts71xxweim changes

Changes made:
- add optional IRQ support
- add get_direction function
- minor misc fixes

These commits have been tested on a ts7250 v3 with FPGA firmware rev70 by jumping two gpios together then using gpiomon on one to listen for interrupts and gpioset on the other to trigger interrupts.

PR #289, opened 2026-03-27
Linux 6.6.y irq fixes

markfeathers:
Added IRQ Ack support
Added GPIO IRQ edge/level support to all TS-7250-V3 GPIO, and first bank of  TS-7100.
Added FPGA rework changes to TS-7100 removing a SPI and adding PWM cores.


ts-kris:
Looks like this supersedes #240 and #275, is that correct?

ts-kris (approved):

markfeathers:
> Looks like this supersedes #240 and #275, is that correct?

Yes, I just closed them with a comment pointing to this.

PR #301, opened 2026-04-24
Input: ads7846 - don't use scratch for tx_buf when clearing register

ts-kris:
Tested on:
- [x] TS-7100-Z (TSC2046)
- [x] TS-TPC-8950-4900 (TSC2046)
- [x] TS-TPC-8390-4900 (ADS7843)

### DIFF
--- a/drivers/input/touchscreen/ads7846.c
+++ b/drivers/input/touchscreen/ads7846.c
@@ -331,7 +331,7 @@ struct ser_req {
 	u8			ref_off;
 	u16			scratch;
 	struct spi_message	msg;
-	struct spi_transfer	xfer[8];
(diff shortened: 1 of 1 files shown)


ts-kris:
Upstreamed patch here: https://lore.kernel.org/lkml/20260424192534.3504976-1-kris@embeddedTS.com/T/#u
v2 patch (fixed up from #303): https://lore.kernel.org/lkml/20260427174657.691272-1-kris@embeddedTS.com/T/#u

Note that this upstream patch should make it back to LTS, and therefore the 6.18 branch here. At some point, this specific commit here will need to be backed out.

markfeathers (approved):

PR #303, opened 2026-04-27
Input: ads7846: fix buffer overrun on clearing command reg

ts-kris:
The previous patch introduced an overflow of scratch on the rx_buf path by reading 3 bytes in to while while scratch is a u16. This shouldn't actually be an issue due to struct alignment, but, the data returned by the command isn't needed anyway so just use NULL on the rx_buf as well.

Fixes: c5257f511bcfd ("Input: ads7846 - don't use scratch for tx_buf when clearing register (#301)")

### DIFF
--- a/drivers/input/touchscreen/ads7846.c
+++ b/drivers/input/touchscreen/ads7846.c
@@ -408,7 +408,6 @@ static int ads7846_read12_ser(struct device *dev, unsigned command)
 	spi_message_add_tail(&req->xfer[5], &req->msg);
 
 	/* clear the command register */
-	req->xfer[6].rx_buf = &req->scratch;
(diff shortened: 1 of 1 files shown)


ts-kris:
Upstreamed patch here: https://lore.kernel.org/lkml/20260427174657.691272-1-kris@embeddedTS.com/T/#u

Note that this upstream patch should make it back to LTS, and therefore the 6.18 branch here. At some point, this specific commit here will need to be backed out.

aaron-ets (approved):

ts-kris:
CI testing keeps failing due to #304, however all other tests pass and this change has no issues when manually tested across all affected platforms.

Issue #304, opened 2026-04-28
linux-6.18.y: i.MX6Q SATA NULL pointer dereference


ts-kris:
There are spurious NULL pointer dereferences with SATA on a TS-8550-4900 under 6.18. The exact cause is currently unknown and its not certain yet whether or not this is a hardware issue with the CI testing rack. This will be further investigated.

```
8<--- cut here ---
Unable to handle kernel NULL pointer dereference at virtual address 00000000 when read
[00000000] *pgd=00000000
Internal error: Oops: 5 [#1] SMP ARM
Modules linked in:
CPU: 3 UID: 0 PID: 65 Comm: scsi_eh_0 Not tainted 6.18.22-g525bbed40a92 #1 PREEMPT 
Hardware name: Freescale i.MX6 Quad/DualLite (Device Tree)
PC is at arch_spin_lock+0x4/0x44
LR is at _raw_spin_lock_irqsave+0x20/0x28
pc : [<c0168aec>]    lr : [<c07754e4>]    psr: 20000093
sp : f0bedd28  ip : 00000000  fp : c054967c
r10: c19f5600  r9 : 10624dd3  r8 : 00000000
r7 : c19f4000  r6 : 00000000  r5 : ffffffa1  r4 : 20000013
r3 : 00000001  r2 : c13c5000  r1 : 00000000  r0 : 00000000
Flags: nzCv  IRQs off  FIQs on  Mode SVC_32  ISA ARM  Segment none
Control: 10c5387d  Table: 1000404a  DAC: 00000051
Register r0 information: NULL pointer
Register r1 information: NULL pointer
Register r2 information: slab task_struct start c13c5000 pointer offset 0 size 2048
Register r3 information: non-paged memory
Register r4 information: non-paged memory
Register r5 information: non-paged memory
Register r6 information: NULL pointer
Register r7 information: non-slab/vmalloc memory
Register r8 information: NULL pointer
Register r9 information: non-paged memory
Register r10 information: non-slab/vmalloc memory
Register r11 information: non-slab/vmalloc memory
Register r12 information: NULL pointer
Process scsi_eh_0 (pid: 65, stack limit = 0x76663fea)
Stack: (0xf0bedd28 to 0xf0bee000)
dd20:                   c19f4000 c0539740 c19f5600 ffffffa1 00000000 c053c9f8
dd40: 00000000 00000000 00000000 c19f5938 c054a724 c054967c 00000000 00000000
[...]

PR #317, opened 2026-05-17
Update to v6.18.32

DaftB0t:
Sync with upstream v6.18.32


ts-kris:
Checks failed due to #304, pushing this PR through for now.

PR #326, opened 2026-06-02
TS-9370 initial support

aaron-ets:
Initial support for TS-9370 i.MX93 board.  Display, camera not supported.


drivers/power/supply/tssilo_supercaps.c
```diff
@@ -0,0 +1,369 @@
...
+					IRQF_ONESHOT, dev_name(dev), data);
+	if (ret)
+		return ret;
+
+	ret = regmap_read(data->regmap, SILO_RESERVED0, &version);
+	if (ret < 0)
+		return ret;
+	dev_info(dev, "TS-SILO version %d\n", version);
+	if (version < 2)
+		dev_warn(dev, "POWER_FAIL ignored without a Wizard interrupt controller.\n");
```
markfeathers: This should be removed, this version register wasn't being updated, and we shouldn't keep using it when its listed as a reserved register.

markfeathers (changes requested):
it looks like 6.18 added new checkpatch updates, or some of these may not have been met originally. For linux-lts though, we generally try to keep that passing with the exception:

DT compatible string "XX" appears un-documented

Those will be fixed when we prep for upstream.

Can you review / fixup the remaining checkpatch warnings?  Eg:
```sh
scripts/checkpatch.pl --git 17cd437bfe05e601ba586d5e370eefa0ffb098aa
```

drivers/power/supply/tssilo_supercaps.c
ts-kris: I'd suggest using `./scripts/checkpatch.pl --git fece377^..HEAD` to run checkpatch against every commit here rather than just the HEAD 17cd437bfe05e601ba586d5e370eefa0ffb098aa. There are a decent amount of warnings and errors that pop up overall. Many can be ignored though, probably.

aaron-ets: The checkpatch warnings/errors I left in:
1. MAINTAINERS file needs to be updated
2. script wanted 4 lines of help text for Kconfig options
3. device tree missing docs
4. There were errors related to wanting () around the pinfunc.h macros, but those are used in the dts file which wouldn't like the ().  I'm surprised checkpatch doesn't have a pincfunc override.

ts-kris: > There were errors related to wanting () around the pinfunc.h macros, but those are used in the dts file which wouldn't like the (). I'm surprised checkpatch doesn't have a pincfunc override.

Yeah, it seems like just something they are dealing with. There may be a day when that comes, but today is not that day.


Everything else you mentioned makes sense and is something to pay closer attention to when upstreaming, case by case.

arch/arm64/boot/dts/freescale/imx93-ts9370-pinfunc.h
ts-kris, line 1: I'm not sure about the organization of this, e.g. should it be in the `freescale/` folder, or our own folder since its IOMUX for our FPGA? should it have `imx93-` as the prefix, since its a pinfunc for our 9370 FPGA that _happens_ to have an i.MX93 CPU?

@markfeathers if you have any ideas of this, let me know. Otherwise I'll probably dig a little bit more just to find any other platforms that might be doing something like this and how they have their IOMUXes laid out.

aaron-ets:
These were mostly just copied as-is from linux-tsimx.  The only changes required were:
1. gpio get/set callbacks return int instead of void
2. structure of ti-adc128s052 driver changed, code to add adc121s021 support was adapted

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+		reg = <0x425e0000 0x10000>;
+		interrupts = <GIC_SPI 55 IRQ_TYPE_LEVEL_HIGH>;
+		clocks = <&clk IMX93_CLK_FLEXSPI1_GATE>,
+		<&clk IMX93_CLK_FLEXSPI1_GATE>;
+		clock-names = "fspi", "fspi_en";
+		assigned-clocks = <&clk IMX93_CLK_FLEXSPI1>;
+		assigned-clock-parents = <&clk IMX93_CLK_SYS_PLL_PFD1>;
+		pinctrl-names = "default";
+		pinctrl-0 = <&pinctrl_flexspi>;
+		//clock-frequency = <66666666>;
```
ts-kris: Remove this comment unless there is a specific reason to leave it as a reference, if so, please add a comment indicating under what circumstances it may be needed.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+			#interrupt-cells = <1>;
+			interrupt-controller;
+		};
+
+		fpga_xbar: pinctrl@28002000 {
+			compatible = "technologic,ts9370-xbar";
+			reg = <0x2000 0x200>;
+		};
+
+		fpga_gpio0: gpio@28000040 {
```
ts-kris, line 420: I'm only familiar with NXPs IOMUX setup, where they have the pinctrl node connected to each gpio instantiation in the devicetree with `gpio-ranges`, e.g.:

```

                        gpio1: gpio@209c000 {
                                compatible = "fsl,imx6ul-gpio", "fsl,imx35-gpio";
...
                                gpio-ranges = <&iomuxc  0 23 10>, <&iomuxc 10 17 6>,
                                              <&iomuxc 16 33 16>;
                        };

                        gpio2: gpio@20a0000 {
                                compatible = "fsl,imx6ul-gpio", "fsl,imx35-gpio";
...
                                gpio-ranges = <&iomuxc 0 49 16>, <&iomuxc 16 111 6>;
                        };
```

Does our driver not need anything like that? Mostly asking just to understand how our driver structure is set up since I otherwise see nothing linking the `fpga_xbar` and individual GPIO controllers. It would make sense if they don't need to be connected in most circumstances.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+	status = "okay";
+
+	mdio {
+		compatible = "snps,dwmac-mdio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		clock-frequency = <5000000>;
+
+		ethphy1: ethernet-phy@1 {
+			//compatible = "ethernet-phy-ieee802.3-c22";
```
ts-kris: Remove commented line unless it makes sense for some reason to leave it in, and if so, add a comment block explaining why it may be useful.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+	fsl,magic-packet;
+	status = "okay";
+
+	mdio {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		clock-frequency = <5000000>;
+
+		ethphy2: ethernet-phy@1 {
+			//compatible = "ethernet-phy-ieee802.3-c22";
```
ts-kris: Same as above

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+	mikrobus_i2c_gpio: i2c-gpio {
+		compatible = "i2c-gpio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		gpios = <
+			&fpga_gpio0 24 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN) /* SDA */
+			&fpga_gpio0 25 (GPIO_ACTIVE_HIGH|GPIO_OPEN_DRAIN) /* SCL */
+		>;
+		i2c-gpio,delay-us = <4>;
+		status = "okay";
```
ts-kris: Not needed when instantiating a new node, default status is `okay`.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+	};
+
+	imx_tac5111: sound-tac5111 {
+		compatible = "simple-audio-card";
+		status = "disabled";
```
ts-kris: I'm not seeing anywhere that this is set to `okay`, is this always non-functional here? Will it be functional later? Given that this is not a dtsi file, I'm assuming there is no plan to have this node be enabled on a platform specific dts?

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+	bluetooth {
+		compatible = "nxp,88w8987-bt";
+		fw-init-baudrate = <3000000>;
+	};
+};
+
+&lpuart7 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_uart7>;
+	linux,rs485-enabled-at-boot-time;
```
ts-kris, line 830: Does this need `uart-has-rtscts` as well to tell the driver to use the hardware RTS pin for RS-485? From the pinmux it looks like we set the pin to UART CTS and RTS mode rather than GPIO.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+	#size-cells = <0>;
+	dr_mode = "host";
+	status = "okay";
+
+	usb@1 {
+		compatible = "usb424,2514";
+		reg = <1>;
+
+		//pinctrl-names = "default";
+		//pinctrl-0 = <&pinctrl_usbhub>;
```
ts-kris: Remove commented line unless it makes sense for some reason to leave it in, and if so, add a comment block explaining why it may be useful.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
```diff
@@ -0,0 +1,1322 @@
...
+&mu1 {
+	status = "okay";
+};
+
+&mu2 {
+	status = "okay";
+};
+
+&ocotp {
+	status = "disabled";
```
ts-kris: No need to access OTP from Linux?

drivers/gpio/Makefile
```diff
@@ -191,6 +191,7 @@ obj-$(CONFIG_GPIO_TQMX86)		+= gpio-tqmx86.o
 obj-$(CONFIG_GPIO_TS4800)		+= gpio-ts4800.o
 obj-$(CONFIG_GPIO_TS4900)		+= gpio-ts4900.o
 obj-$(CONFIG_GPIO_TS5500)		+= gpio-ts5500.o
+obj-$(CONFIG_GPIO_TS9370)		+= gpio-ts9370.o
 obj-$(CONFIG_GPIO_TS71XXWEIM)		+= gpio-ts71xxweim.o
 obj-$(CONFIG_GPIO_TS7800V2)		+= gpio-ts7800-v2.o
```
ts-kris:
```suggestion
obj-$(CONFIG_GPIO_TS71XXWEIM)		+= gpio-ts71xxweim.o
obj-$(CONFIG_GPIO_TS7800V2)		+= gpio-ts7800-v2.o
obj-$(CONFIG_GPIO_TS9370)		+= gpio-ts9370.o
```

Keep these listed in numerical order for consistency.

drivers/irqchip/Makefile
```diff
@@ -62,6 +62,8 @@ obj-$(CONFIG_ARCH_VT8500)		+= irq-vt8500.o
 obj-$(CONFIG_ST_IRQCHIP)		+= irq-st.o
 obj-$(CONFIG_TB10X_IRQC)		+= irq-tb10x.o
 obj-$(CONFIG_TS4800_IRQ)		+= irq-ts4800.o
+obj-$(CONFIG_TS9370_IRQ)		+= irq-ts9370.o
+obj-$(CONFIG_WIZARD_IRQ)		+= irq-wizard.o
 obj-$(CONFIG_TSWEIM_FPGA_INTC)		+= irq-ts71xxweim.o
```
ts-kris, line 66:
```suggestion
obj-$(CONFIG_TS9370_IRQ)		+= irq-ts9370.o
obj-$(CONFIG_TSWEIM_FPGA_INTC)		+= irq-ts71xxweim.o
obj-$(CONFIG_WIZARD_IRQ)		+= irq-wizard.o
```

Reorder this to be in numeric/alphabetical order for consistency.

drivers/irqchip/irq-wizard.c
```diff
@@ -0,0 +1,94 @@
...
+	REGMAP_IRQ_REG(7, 0, BIT(7)),
+	REGMAP_IRQ_REG(8, 0, BIT(8)),
+	REGMAP_IRQ_REG(9, 0, BIT(9)),
+	REGMAP_IRQ_REG(10, 0, BIT(10)),
+	REGMAP_IRQ_REG(11, 0, BIT(11)),
+	REGMAP_IRQ_REG(12, 0, BIT(12)),
+	REGMAP_IRQ_REG(13, 0, BIT(13)),
+	REGMAP_IRQ_REG(14, 0, BIT(14)),
+	REGMAP_IRQ_REG(15, 0, BIT(15)),
+};
```
ts-kris, line 39: From what I can tell, the preferred way of handling this, with the offset always being the same (`0`), is `REGMAP_IRQ_REG_LINE()` from 43fac3238c1d9363b2a93d8d56c2be0c29c64e6c

```        
#define REGMAP_IRQ_REG(_irq, _off, _mask)               \
        [_irq] = { .reg_offset = (_off), .mask = (_mask) }
        
#define REGMAP_IRQ_REG_LINE(_id, _reg_bits) \
        [_id] = {                               \
                .mask = BIT((_id) % (_reg_bits)),       \
                .reg_offset = (_id) / (_reg_bits),      \
        }
```

```suggestion
#define MAX_IRQS	16

#define IRQ_STATUS      0
#define IRQ_ACK         1
#define IRQ_MASK_SET    2
#define IRQ_MASK_CLR    3
#define IRQ_MASK_RW     4

static const struct regmap_irq wizard_irqs[] = {
	REGMAP_IRQ_REG_LINE(0, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(1, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(2, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(3, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(4, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(5, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(6, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(7, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(8, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(9, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(10, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(11, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(12, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(13, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(14, MAX_IRQS),
	REGMAP_IRQ_REG_LINE(15, MAX_IRQS),
}
```


The former in our case producing, e.g.
```
[0] = { .reg_offset = 0, .mask = (1 << 0) },
[1] = { .reg_offset = 0, .mask = (1 << 1) },
[2] = { .reg_offset = 0, .mask = (1 << 2) },
```

And the latter producing (reordered for consistency):
```
[0] = { .reg_offset = ((0/16) == 0), .mask = (((1 << 0) % 16) == 1) }
[1] = { .reg_offset = ((1/16) == 0), .mask = (((1 << 1) % 16) == 2) }
[2] = { .reg_offset = ((2/16) == 0), .mask = (((1 << 2) % 16) == 4) }
```

This may not be worth changing for this PR, but may need to be kept in mind when upstreaming later.

arch/arm64/configs/tsimx93_defconfig
ts-kris, line 1: It looks like our defconfig defaults to having `SOC_IMX8` as well as a handful of other default enabled i.MX8 drivers like clocks etc that I don't think we need? Unless the i.MX9 uses those under the hood. Would potentially benefit us to do a sweep and disable things like that which we don't need for sure to save some space.

drivers/pinctrl/technologic/Kconfig
ts-kris, line 1: I don't have a better answer for this right now, but I'm not a fan of having separate selections for both platforms. It might be cleaner to have a single config option that is per CPU series, since in most cases they would both/all be enabled in our defconfigs, and it would require tuning for specific platforms if a customer wanted that.

It may also be a bright idea to completely hide these (or at least have some reverse dependency saying that if the ts9370 GPIO driver was selected, then select the pinmux driver too).

drivers/power/supply/tssilo_supercaps.c
```diff
@@ -0,0 +1,361 @@
...
+static int ts_silo_probe(struct platform_device *pdev)
+{
+	struct ts_wizard *wizard = dev_get_drvdata(pdev->dev.parent);
+	struct device *dev = &pdev->dev;
+	struct silo_data *data;
+	struct power_supply_config psy_cfg = {};
+	int ret;
+	int irq;
+
+	data = devm_kzalloc(dev, sizeof(*data), GFP_KERNEL);
```
ts-kris, line 310: Return of `devm_kzalloc()` is unchecked

ts-kris (changes requested):
A couple of small fixes, couple of comments that would be good to talk about.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: This was disabled in @markfeathers  commit 08074a5eeea4c9ce677d181d1bcf14da0a2d4d7d on linux-tsimx with commit log:

```
arm64: dts: imx93-ts9370: Add missing fuse block, add wdog pinctrl
```

Does the ocotp conflict with the wdog?

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: I don't think you need uart-has-rtscts, I think by default it uses the RTS line as RE/DE in rs485 mode.  It definitely worked in tsimx kernel.  I would assume we don't need the CTS line in pinctrl, I'll try it without.

arch/arm64/configs/tsimx93_defconfig
aaron-ets: SOC_IMX8M and SOC_IMX9 are both set by `default ARCH_MXC && ARM64`.  We could roll this sweep into the defconfig cleanup task, does not seem like these defconfig files will be upstreamable anyway.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: After talking to @markfeathers it seems like rs485 doesn't need CTS or uart-has-rtscts.  However, it would be nice if user could switch it from rs485 to rs232 with an ioctl at runtime without needing a device tree overlay, in which case it would need both to support hardware flow control.

drivers/pinctrl/technologic/Kconfig
aaron-ets: I could just remove 9390 for the moment and punt until we have one to test, I didn't copy over the 9390 device trees anyway.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
ts-kris: Ah, looks like `uart-has-rtscts;` is driver dependent, some drivers gate updating CTS/RTS bits in the peripheral based on this flag. I forgot the i.MX9 uses `fsl_lpuart.c` not the generic `imx.c` which _does_ require `uart-has-rtscts;` for doing anything with the real hardware RTC/CTS.

>  However, it would be nice if user could switch it from rs485 to rs232 with an ioctl at runtime without needing a device tree overlay, in which case it would need both to support hardware flow control.

Looked a bit closer at the driver as it exists, and, yeah, not a path that already exists to handle that automatically. Though we can do it without a devicetree overlay/change.

If we are defaulting to RS-485 mode with `linux,rs485-enabled-at-boot-time;` then we can default to de-asserting `EN_2ND_PORT_232` GPIO as well. To use it in RS-232, ioctl() can clear RS485 on the port, and then assert `EN_2ND_PORT_232` to enable RX drivers on RS-485 transceiver.

The safest state would probably be to not have `linux,rs485-enabled-at-boot-time;`, and then `EN_2ND_PORT_232` asserted. Then to enable RS-485, deassert the `EN_2ND_PORT_232` GPIO, and then its safe to open the port and set the 485 ioctl() on it.

arch/arm64/configs/tsimx93_defconfig
ts-kris: Correct, which is why for us, it would make sense to trim that out to save on unnecessary additional dead code. I'm find if you open an Issue and we kick it down the road a bit, if you prefer.

drivers/pinctrl/technologic/Kconfig
ts-kris: I don't have a preference, nor an answer, but it should be something we look at closer before we upstream this.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Removed

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Removed

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Removed

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Removed

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: The audio card wasn't working and I thought Mark had said the audio interface wasn't working on the i.mx93 upstream yet.  But looking at the imx93 evks they have audio enabled.  Issue was missing TI codec driver, added it in new commit, audio is working as far as audio device showing up without error now.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Mark was referring more to routing the uart7 pins to the daughter card pins through the FPGA and avoiding the transceiver.  But maybe just default to only rs485 and require an overlay to change it would be ok.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Removed

drivers/gpio/Makefile
aaron-ets: Updated

drivers/irqchip/Makefile
aaron-ets: Udpated.

drivers/power/supply/tssilo_supercaps.c
aaron-ets: Updated to return ENOMEM

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
ts-kris: Ah, okay. Then yeah, thats gonna be an overlay for sure since that would require an iomux change.

aaron-ets: So talked to Mark and Joshua about this.  The problem with lpuart2 232 at boot time and switching to 485 at runtime is that the RX line is driven high in 232.  If there are 485 devices connected to the 9370 it would interfere on the bus during boot.  If we default to 485 at boot and there is a 232 device connected should not cause any problems, with send disabled.  However, we need a hardware pulldown on `EN_2ND_PORT_232` in the next revision to really cover the case at boot.  In any case for the moment seems safest to have lpuart3 and lpuart7 enable rs485 at boot.

arch/arm64/configs/tsimx93_defconfig
markfeathers: We were forced to enable them on the nxp kernel fork because many of the forked nxp drivers are not buildable if you disable some of those options. I'd expect we are able to trim this a bit more safely on the LTS kernels.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
ts-kris: Yeah, that sounds good to me. Sounds like nothing else doing on this point, resolving.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Mark wasn't sure why ocotp was disabled.  Re-enabled it and wdt was still working.

arch/arm64/boot/dts/freescale/imx93-ts9370-pinfunc.h
markfeathers: When boards need header files they are typically in the same path, but perhaps its wrong to have imx93- in the front?  I'm not sure if there is a more normal pattern for this from other devices.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
markfeathers: I think from our discussion we decided we do not want to connect this at this time, but we can wire it up later if needed.

drivers/irqchip/irq-wizard.c
markfeathers: This looks simpler to me, I'd agree we should just change this.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
ts-kris: This looks like its still present in the file. I can't tell if it got added back in, or if git didn't update the files correctly or what.

arch/arm64/configs/tsimx93_defconfig
ts-kris: Going to be addressed in #341

ts-kris (commented):
See the one open comment still about `status` on a new devicetree node.

arch/arm64/boot/dts/freescale/imx93-ts9370.dts
aaron-ets: Looks like it was not removed and I lied about it.  Let me update it and also rebase on latest linux-6.18.y and retest real quick.

aaron-ets: Actually Removed, probably

ts-kris (approved):

markfeathers (approved):

PR #330, opened 2026-06-19
Update to v6.18.36

DaftB0t:
Sync with upstream v6.18.36


ts-kris:
Pipeline failing due to #304, forcing merge.

PR #339, opened 2026-07-21
Max3100 ts fix

aaron-ets:


drivers/tty/serial/max3100-ts.c
ts-kris:
```suggestion
	struct max3100_port *s = NULL;
```

No need to get the drvdata here, we're not using it. However, `*s` is used below so we still need to declare it.

drivers/tty/serial/max3100-ts.c
```diff
@@ -938,12 +936,10 @@ static void max3100_remove(struct spi_device *spi)
 
 static int max3100_suspend(struct device *dev)
 {
-	struct max3100_port *s = dev_get_drvdata(dev);
+	struct max3100_port *s = 0;
```
ts-kris:
```suggestion
	struct max3100_port *s = NULL;
```

Since `NULL` is usually defined as `((void *)0)`, `NULL` is more complete here and won't error on strict compilers.

drivers/tty/serial/max3100-ts.c
```diff
@@ -938,12 +936,10 @@ static void max3100_remove(struct spi_device *spi)
...
 	u16 rx;
 	int i;
 
-	dev_dbg(&s->spi->dev, "%s\n", __func__);
```
ts-kris, line 945: I think it would be worth moving this inside the loop below after `*s` is assigned. So there is still a debug print for each port that is being suspended. Rather than just silence.

I think, instead, this may be able to just use the same `dev` that is passed to this function?

drivers/tty/serial/max3100-ts.c
```diff
@@ -963,11 +959,9 @@ static int max3100_suspend(struct device *dev)
 
 static int max3100_resume(struct device *dev)
 {
-	struct max3100_port *s = dev_get_drvdata(dev);
+	struct max3100_port *s = 0;
 	int i;
 
-	dev_dbg(&s->spi->dev, "%s\n", __func__);
```
ts-kris, line 969: I think it would be worth moving this inside the loop below after `*s` is assigned. So there is still a debug print for each port that is being suspended. Rather than just silence.

I think, instead, this may be able to just use the same `dev` that is passed to this function?

drivers/tty/serial/max3100-ts.c
```diff
@@ -963,11 +959,9 @@ static int max3100_suspend(struct device *dev)
 
 static int max3100_resume(struct device *dev)
 {
-	struct max3100_port *s = dev_get_drvdata(dev);
+	struct max3100_port *s = 0;
```
ts-kris:
```suggestion
	struct max3100_port *s = NULL;
```

Since `NULL` is usually defined as `((void *)0)`, `NULL` is more complete here and won't error on strict compilers.

ts-kris (changes requested):

ts-kris (approved):

